PR #1, opened 2023-11-27
UPDATE: latest docs onto this new structure

daquinteroflex:


daquinteroflex:
So the idea of this pull request is that it is a live checker that I've properly updated all the notebooks. I will manually attempt to merge this on the develop branch, and then this PR verifies that there are no conflicts missing or things missing from the integration when evaluating the changes section.

PR #12, opened 2024-01-02
Added three optimization notebooks and associated files

tomflexcompute:
Kicking off the new year with a larger notebook addition. This PR contains three new case studies that use different gradient-free optimization algorithms that are popular in photonic design, in addition to the particle swarm optimization (PSO) case studies that we already have. After this, our library should've covered most commonly used optimization techniques, which hopefully can stimulate more Tidy3D usage. 

The three notebooks are:

- Direct binary search (DBS) for an optical switch.
- Genetic algorithm (GA) for a reflector.
- CMA-ES for a waveguide s-bend.

Having more people review here since it contains more content. Nothing urgent so feel free to take your time. @daquinteroflex could you check if the PR doesn't contain any mistakes or miss anything as it's the first notebook submission in the new repo. I think the workflow is similar but good to have you double-check. 

As a side note, interestingly in both the GA and DBS, the results I got are even a bit better than what's shown in the referenced paper, with a smaller number of simulation runs. I think this is due to a combination of using symmetry and luck (the random seed I used might have been a bit better by chance). Not sure why the papers don't use symmetry.


GeneticAlgorithmReflector.ipynb
tylerflex, line 1: Consider a `GeometryGroup` for grouping all the pixels in a single `Structure`. This will make the plotting and processing faster (not sure if it's currently an issue, but could be good to highlight this usage if users want to extend this example).

tylerflex (changes requested):
This is really cool! You were busy over the holidays I guess? ;).

I just looked at the GA one, I'm excited to see this as I enjoyed playing with these algorithms many years ago before I got into adjoint work.

Only one comment there for now. I'll take a look at the other ones later today.

GeneticAlgorithmReflector.ipynb
tomflexcompute: Thanks. At the moment the performance isn't really a problem since the number of structures is still relatively small. But that's a good idea. I'll revise the notebooks to demonstrate the best practice. Will wait for others' comments and do a larger revision later on as running the notebooks takes a bit of time (45-60 mins). 

My wife and I don't really celebrate chrismas but we will probably take some time off for the Chinese new year, which is later in February this year.

tomflexcompute:
By the way in the original docs repo there is a requirements.txt file. In this repo the package list in the requirements.txt is much shorter. Things like `pyswarms` and `sax` are removed. Do we still need them in requirements.txt as some notebooks import them? @daquinteroflex

OpticalSwitchDBS.ipynb
tylerflex, line 1: Looks good!

1. Might be interesting to see the objective function vs iteration plot.
2. There's also a variant of this called "simulated annealing", where the flip to a worse device can happen with some probability (controlled by a "temperature"). This probability is initially high but is reduced as the simulation is "cooled" down (like annealing). It could be interesting to try this in the future.

SbendCMAES.ipynb
tylerflex, line 1: 1. "a S-bend -> an S-bend"
2. This is an interesting approach and seems even better than adjoint with curved surfaces (eg. bend radius) in terms of performance improvement / iteration. Is there a downside to this approach vs adjoint? maybe less flexible? how does one enforce feature size constraints (eg removing small bend radius)? is it possible or does it need to be built into the parameterization?

tylerflex (commented):
just FYI make sure you add these to the `examples.rst` which now lives on the front end repo. Probably makes sense to organize the adjoint and optimization into separate subsections of an "optimization" section?

SbendCMAES.ipynb
tomflexcompute: The downside seems to be just more simulations need to be run in general. I think the flexibility in terms of possible shapes can be increased by using more control points in the Bezier curve. The number of control points should also influence the possible curvature. Increasing the number of support points will likely decrease the minimal curvature. One way to ensure sufficiently large curvature would be to constrain the minimal distance between two support points. It's likely not a concern for a small number of points like this case tho.

tomflexcompute:
> just FYI make sure you add these to the `examples.rst` which now lives on the front end repo. Probably makes sense to organize the adjoint and optimization into separate subsections of an "optimization" section?

Yeah @e-g-melo and I were thinking about it too. Currently some adjoint notebooks are missing thumbnails for the marketing site. After this we can make a separate PR for reorganizing all optimization notebooks.

Sorry where is the `examples.rst`? I didn't find it in the frontend repo either.

GeneticAlgorithmReflector.ipynb
e-g-melo, line 1: Hi Tom!

1) It is a pretty good GA example.
a) It is interesting to see that you have obtained a good convergence even using a population of only 30 individuals. In most of the GA papers I saw, they set a population of at least 7*Nx*Ny.
b) I wonder how long it takes to run the entire optimization.

OpticalSwitchDBS.ipynb
e-g-melo, line 1: Very good as well!

1. Before [12]: "very mildly the the third iteration" -> "very mildly the third iteration"
2. Before [16]: "Finally plot the electric field" -> "Finally, plot the electric field" 
3. I agree with Tyler, the objective function x iteration plot would be nice.

SbendCMAES.ipynb
e-g-melo, line 1: 1. At introduction:

```
1. Initialization: Generate an initial population of candidate solutions randomly within the search space.

Initialize the parameters: mean of the search distribution, covariance matrix, step size, etc.
```
Maybe?
```
1. Initialization: Generate an initial population of candidate solutions randomly within the search space. Initialize the parameters: mean of the search distribution, covariance matrix, step size, etc.
```

2. Before [9]: "returns a 1D array of objective functions" -> "returns a 1D array of objective function values"?
3. Following previous discussions, it seems you have defined the objective function as the insertion loss in [9] (`T.append(-10 * np.log10(np.mean(compute_transmission(sim_data))))`) and the final result as the transmission in [15] (`T_opt = 10 * np.log10(compute_transmission(sim_data_opt))`). Maybe use insertion loss in [15] as well?

e-g-melo (approved):
Pretty good examples Tom! 

I suggest you create a new section to group all the non-gradient-based optimization examples so the users can find them easily.

tomflexcompute:
> Pretty good examples Tom!
> 
> I suggest you create a new section to group all the non-gradient-based optimization examples so the users can find them easily.

Thanks Emerson! Will implement these changes. Indeed the papers used a population of 200. I think I got away with a smaller population due to the use of symmetry and a bit of luck (better random seed and hyperparameters).

daquinteroflex (approved):
Yeah this follows the new structure well, and think it should be good. Have you built locally using `tidy3d develop build-docs` on how it looks like if you wanted to verify the whole flow integration?

I have one slightly related question. On the old workflow, did you manually "sync-to-proxy" repo to update the notebooks to MC server or did you expect it to occur automatically?

tomflexcompute:
> > just FYI make sure you add these to the `examples.rst` which now lives on the front end repo. Probably makes sense to organize the adjoint and optimization into separate subsections of an "optimization" section?
> 
> Yeah @e-g-melo and I were thinking about it too. Currently some adjoint notebooks are missing thumbnails for the marketing site. After this we can make a separate PR for reorganizing all optimization notebooks.
> 
> Sorry where is the `examples.rst`? I didn't find it in the frontend repo either.

Are you referring to [this](https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/develop/docs/source/examples.rst)?

tomflexcompute:
> Yeah this follows the new structure well, and think it should be good. Have you built locally using `tidy3d develop build-docs` on how it looks like if you wanted to verify the whole flow integration?
> 
> I have one slightly related question. On the old workflow, did you manually "sync-to-proxy" repo to update the notebooks to MC server or did you expect it to occur automatically?

Thanks @daquinteroflex for taking a look. I can try to build it locally. 

The update occurs automatically and is directly triggered by the commit I think.

daquinteroflex:
Sounds good, I think I have to enable the automatic updates then - was unsure if we intended manual updates. Let me know if you need any help building the docs or if you prefer I do it instead in this case (although it's good practice if you ever want to edit it in the future) too

daquinteroflex:
> just FYI make sure you add these to the `examples.rst` which now lives on the front end repo. Probably makes sense to organize the adjoint and optimization into separate subsections of an "optimization" section?

Actually `examples.rst` is no longer active. It got broken down into the `tidy3d-noteboks/docs` directory which I think Tom edited well and I think should work

tylerflex:
>> just FYI make sure you add these to the examples.rst which now lives on the front end repo. Probably makes sense to organize the adjoint and optimization into separate subsections of an "optimization" section?

> Actually examples.rst is no longer active. It got broken down into the tidy3d-noteboks/docs directory which I think Tom edited well and I think should work

Oh, sorry about that, I forgot how that worked, thanks!

GeneticAlgorithmReflector.ipynb
tomflexcompute: Turns out that the way I currently have is already using the GeometryGroup. When adding two geometries, the result is automatically a GeometryGroup.

PR #13, opened 2024-01-02
Adjoint quickstart notebook

tylerflex:
Adds a quickstart example for getting adjoint running. We probably should recommend people reference this notebook initially when introducing adjoint.

Also adds a link to the quickstart from the original Intro notebook.

Adding @daquinteroflex mainly to make sure I set the PR up properly, but feel free to look at the notebook too if you're curious.


daquinteroflex (approved):
Looks well set up! And also quite interesting - nice that it was boiled down to just the essential gradient function optimizations.

For someone skimming or just getting also started with tidy3d a little illustration of the simulation problem might help them conceptualise the code before they reach the end. It's pretty clear and simple anyway anwyay just thought maybe a nice little addittion if we wanted? Like a little point showing the point source and measurement, and arrows on a box dimension side showing with a gradient function for the area might help illustrate it for someone looking through all the notbeooks and want to get an idea what's in it.

tomflexcompute (approved):
Thanks for the great adjoint intro demo. Since it's a 101 kind of demo, maybe it's worth it to briefly introduce the adjoint method and discuss the key benefit in the very beginning? Something like "Using the adjoint method, you can compute the gradient of the objective function with respect to an arbitrary number of design parameters in just two simulations, which is far more computationally efficient than other optimization algorithms." Just so that users without any adjoint knowledge can get a sense of the benefit.

In addition, "Now we put construct our objective function" -> "Now we construct our objective function".

Following Dario's comment, maybe it's worth it to put a simple thumbnail image to quickly demonstrate the idea and simulation setup? I can also prepare a 3D rendering if you feel like it can be helpful.

tylerflex:
I changed the intro to the following text
```
This notebook will get users up and running with a very simple inverse design optimization with `tidy3d`. Inverse design uses the "adjoint method" to compute gradients of a figure of merit with respect to design parameters using only 2 simulations no matter how many design parameters are present. This gradient is then used to do high dimensional, gradient-based optimization of the system.

The setup we'll demonstrate here involves a point dipole source and a point field monitor on either side of a dielectric box. Using the adjoint plugin in `tidy3d`, we use gradient-based optimization to maximize the intensity enhancement at the measurement spot with respect to the box size in all 3 dimensions.
```

And added this image
image

If this looks good feel free to merge this pull request or let me know so I can merge it. Thanks!

daquinteroflex:
Looks good feel free to merge!

tomflexcompute:
Thanks for the update. Looks great!

tylerflex:
I messed up and there was a small bug (the field monitor was moved to the end) in the notebook I ended up sending, re-running now and will make a new PR

PR #14, opened 2024-01-05
Demonstration of Simulation.subsection in thermal ring resonator

dbochkov-flexcompute:
A small modification to thermal ring resonator notebook to demonstrate one possible usage of `Simulation.subsection` (just to remove structures in PML and outside) + switch mode solver calls in there to remote one because now we can do that without uploading/downloading 500mb of data.


daquinteroflex:
@dbochkov-flexcompute  would you like me to sort out the git history of your branch?

PR #16, opened 2024-01-06
Fabricability improvement at interface.

e-g-melo:
Inclusion of a buffer layer to improve fabricability at the design region and connection interface.


tylerflex (approved):
looks great, thanks!

PR #24, opened 2024-01-11
Explanations for mode solver simulation reduction in heat notebooks

dbochkov-flexcompute:


momchil-flex (approved):
Looks good, but do you want to just include this in #26 ?

dbochkov-flexcompute:
yeah, I can consolidate

PR #25, opened 2024-01-11
Carrier depletion modulator example

dbochkov-flexcompute:
Example using open-source package devsim as a charge solver


tomflexcompute (commented):
Thanks @dbochkov-flexcompute for the great efforts! We can share this demo with a few current customers who are interested in electro-optic modulators after it's published. Some comments I have at the moment:

- I think you also need to modify the rst file in docs/case_studies accordingly depending on if you want to add this notebook to a current category or create a new category. 
- Consider adding a link to the devsim docs or repo in the introduction so users can see how to install it and find relevant documentation. 
- The carrier distribution simulation using devsim seems a bit involved, including mesh creation, parameters setting, etc. Could you add a bit more description in the markdown on each step? It could help users understand the workflow better. Potentially consider adding references to devsim's relevant examples in their docs if there are any.
- Usually it's more customary and intuitive to plot the spectral response in wavelength instead of frequency.

Some typos I caught:
- "due to presence of free carriers" -> "due to the presence of free carriers"
- "one of which is has a common pin-junction profile" -> "one of which has a common pin-junction profile"
- "paramters" -> "parameters"
- "heavy doped" -> "heavily doped"
- "we will expanded device" -> "we will expand device"
- "to define other component of an optic simulation" -> "to define other components of an optic simulation"
- "no electron and hole distribution have been provided to it" -> "no electron and hole distributions have been provided to it"

Regarding the final result that the transmission doesn't change much with respect to the applied voltage, I suppose it's not surprising as the real modulator arms can be mm in length. Nontheless it's still a good demonstration. However in a more practical sense, I suppose people usually only simulate a small portion of the modulated arm only without the couplers and splitters and then extrapolate the modulation result to the full length. Maybe we can provide another simpler notebook for that. It's similar to the [thermally tuned waveguide](https://www.flexcompute.com/tidy3d/examples/notebooks/MetalHeaterPhaseShifter/) that you made.

e-g-melo (approved):
Hi @dbochkov-flexcompute! Thanks for this excellent example!

1- Include the units in doping concentrations in [2].
2 - It is not so easy to observe the changes in depletion region in the figure in [6]. Maybe limit the y-axis between 2 and 3? The same occurs in [39].
3 - Include the units in x, y, z-axis.
4 - I suggest to include vertical lines to separate all those constants before [8].
5 - It would be nice to include some comments to tell the users that this example illustrates how to use `ChargeDataArray`, `CustomChargePerturbation`, and `ParameterPerturbation` (including links to API). That should also make it easier for our search engine to find it.
6 - Are we missing a link here? "For generating the entire circuit we use helper functions for creating a single waveguide and a single coupler from tutorial ."
7 - It seems the `Scene` object is duplicated in [16] and [17].
8 - I suggest to include a legend in figure [32].

Other typo:
"data from charge solver to opitc solver" -> "data from charge solver to optic solver"

I wonder how this example would be when using the future Charge solver. Would it simplify that long function to obtain the carrier distribution?

dbochkov-flexcompute:
Taking your comments, and doing some more research on available literature, I made quite an overhaul of this example. Here are major changes:
- switched from carrier depletion (reverse bias) to carrier injection (forward bias) operation mode. This is because I found some experimental results in one of the papers for a 200 um long modulator that we could qualitatively compare to. The overall demonstration becomes more obvious too, since the propagation index change is more pronounced.
- change splitter and combiner to y junction to reduce simulation size
- reorganized convenience functions for creating components


Things that I couldn't address or will address in separate PRs:

@e-g-melo:

> 2 - It is not so easy to observe the changes in depletion region in the figure in [6]. Maybe limit the y-axis between 2 and 3? The same occurs in [39].

now with introduced changes I believe this is not relevant

> 4 - I suggest to include vertical lines to separate all those constants before [8].

unfortunately, it seems like how tables are rendered is not controlled by markdown code, but the html style

> 5 - It would be nice to include some comments to tell the users that this example illustrates how to use ChargeDataArray, CustomChargePerturbation, and ParameterPerturbation (including links to API). That should also make it easier for our search engine to find it.

actually, there is a forgotten notebook https://github.com/flexcompute-readthedocs/tidy3d-docs/blob/4bdbc405ea07606b46dbe575798ecede298fcf3e/docs/source/notebooks/PerturbationMedium.ipynb that explains this in detail. I will refresh it and submit in a separate RP (with adding a link into this PR's notebook as well)


Other comments:

> I wonder how this example would be when using the future Charge solver. Would it simplify that long function to obtain the carrier distribution?

@e-g-melo, yeah, it would be very similar to our heat solver API

e-g-melo:
Hi @dbochkov-flexcompute! Thanks for your answers!

The final result has improved a lot! It is much better and easier to observe the changes in permittivity now.

tomflexcompute:
Thanks @dbochkov-flexcompute for the update. The current design is indeed very nice. 

For notebooks under case studies, I think we do need a thumbnail image for the [marketing site](https://www.flexcompute.com/tidy3d/learning-center/example-library/). I can help with a 3D schematic later today. 

For the electric field plots, can we set the aspect ratio to `auto` as the aspect ratio of the MZI is quite large. Using an equal aspect ratio makes details a bit hard to see.

In the last plot, there are noticeable fluctuations in the transmission. Is that physical or due to the grid size not sufficiently fine?

dbochkov-flexcompute:
Realized that doping concentrations were not set correctly, fixed that. This brought the predicted Vpi from 1V down to about 0.95 V.

> In the last plot, there are noticeable fluctuations in the transmission. Is that physical or due to the grid size not sufficiently fine?

@tomflexcompute 
it seems to be physical, probably due to some transitions are not gradual enough?

Screenshot_20240124_145447

tomflexcompute:
I see. Yeah that could be the case due to some reflection and interference, which is fine. 

Btw I tried to run this notebook after installing devsim but the kernel kept dying during the charge solving step. Not sure why so just wondering if you have experienced the same thing before?

dbochkov-flexcompute:
> I see. Yeah that could be the case due to some reflection and interference, which is fine.
> 
> Btw I tried to run this notebook after installing devsim but the kernel kept dying during the charge solving step. Not sure why so just wondering if you have experienced the same thing before?

hm, maybe running out of RAM. Can you try setting lower resolution for solving for charge? `res = 0.005` in cell 5

tomflexcompute:
> > I see. Yeah that could be the case due to some reflection and interference, which is fine.
> > Btw I tried to run this notebook after installing devsim but the kernel kept dying during the charge solving step. Not sure why so just wondering if you have experienced the same thing before?
> 
> hm, maybe running out of RAM. Can you try setting lower resolution for solving for charge? `res = 0.005` in cell 5

Actually the kernel dies when trying to plot the results. I tried `res=0.05` and the simulation is very fast but the kernel still dies when plotting. Maybe it's my computer's issue.

dbochkov-flexcompute:
> > > I see. Yeah that could be the case due to some reflection and interference, which is fine.
> > > Btw I tried to run this notebook after installing devsim but the kernel kept dying during the charge solving step. Not sure why so just wondering if you have experienced the same thing before?
> > 
> > 
> > hm, maybe running out of RAM. Can you try setting lower resolution for solving for charge? `res = 0.005` in cell 5
> 
> Actually the kernel dies when trying to plot the results. I tried `res=0.05` and the simulation is very fast but the kernel still dies when plotting. Maybe it's my computer's issue.

is `vtk` installed in your environment?

tomflexcompute:
> > > > I see. Yeah that could be the case due to some reflection and interference, which is fine.
> > > > Btw I tried to run this notebook after installing devsim but the kernel kept dying during the charge solving step. Not sure why so just wondering if you have experienced the same thing before?
> > > 
> > > 
> > > hm, maybe running out of RAM. Can you try setting lower resolution for solving for charge? `res = 0.005` in cell 5
> > 
> > 
> > Actually the kernel dies when trying to plot the results. I tried `res=0.05` and the simulation is very fast but the kernel still dies when plotting. Maybe it's my computer's issue.
> 
> is `vtk` installed in your environment?

Yeah it's probably one of these weird things that happen out of nothing. It should be fine. 

Do you think we can use this image as the thumbnail?
image

dbochkov-flexcompute:
This looks great! Do I need to add it somewhere?

tomflexcompute:
> This looks great! Do I need to add it somewhere?

I'll push something to take care of it. By the same token we can use this image for the thermally tuned ring resonator. Do you want to move that notebook under the PIC category too since I see this MZI is under PIC already?


image

tomflexcompute:
Ok I've included the two thumbnail images in the notebooks so they can be displayed in our [example library](https://www.flexcompute.com/tidy3d/learning-center/example-library/) once they are released. Feel free to decide if you want to move the ring resonator to PIC or leave it as a heat solver tutorial. Then I think this PR is ready. In case @e-g-melo wants to take another look?

dbochkov-flexcompute:
@tomflexcompute thanks for taking care of that! those are very nice illustrations. 

Regarding example categories: maybe we should create a separate category "Active Photonic Integrated Circuit Components", which would include for now:
- Thermally tuned waveguide
- Thermally tuned ring resonator
- Carrier injection based Mach-Zehnder modulator

what do you think?

tomflexcompute:
> @tomflexcompute thanks for taking care of that! those are very nice illustrations.
> 
> Regarding example categories: maybe we should create a separate category "Active Photonic Integrated Circuit Components", which would include for now:
> 
> * Thermally tuned waveguide
> * Thermally tuned ring resonator
> * Carrier injection based Mach-Zehnder modulator
> 
> what do you think?

Yeah that certainly sounds like a great idea! Can also change the current "Photonic Integrated Circuit Components" to "Passive Photonic Integrated Circuit Components".

PR #26, opened 2024-01-12
A few corrections in heat solver notebooks

dbochkov-flexcompute:
Same as https://github.com/flexcompute-readthedocs/tidy3d-docs/pull/346


tomflexcompute:
Thanks @dbochkov-flexcompute . I think it can go into `develop`? Was the other PR in tidy3d-docs merged into the pre branch?

momchil-flex:
Wait we shouldn't be putting things in develop in this repo. This repo is only used starting from 2.6.

dbochkov-flexcompute:
@momchil-flex Yeah, I am not sure what's the best, I added these corrections in 2.5 docs https://github.com/flexcompute-readthedocs/tidy3d-docs/pull/346, but I imagine propagating them from there to 2.6 would results in some annoying conflicts, so I was thinking of introducing the same changes in 2.6 docs directly so we can just overwrite these notebooks. Either merging this into develop or pre/2.6 is fine with me, whatever makes sense

momchil-flex:
Yeah so currently my understanding is that we won't be merging anything from tidy3d-docs here. Instead in the transition period we need to make sure any changes in tidy3d-docs/develop are also put here in pre/2.6.

PR #27, opened 2024-01-22
Modified the bragg gratings notebook.

tomflexcompute:
- Modified the bragg gratings notebook to address https://github.com/flexcompute/tidy3d/issues/1339. 

- Small Formatting fix for the chiral metasurface and y splitter notebooks along the way.


e-g-melo (approved):
Hi @tomflexcompute! Thanks for improving these examples.

The Bragg waveguides example is much better now with the structures extending into the PML. The results are very clear without those Fabry-Perot fringes. The other ones have only small formating changes, right?

daquinteroflex:
Looks good and also properly implemented! 

When you do these updates I assume you want it to automatically trigger a new docs build? Currently you have to update the submodule state on pre/2.6 in the main repo. However, I can write an action to trigger this automatically if you need.

daquinteroflex (approved):

tomflexcompute:
> Hi @tomflexcompute! Thanks for improving these examples.
> 
> The Bragg waveguides example is much better now with the structures extending into the PML. The results are very clear without those Fabry-Perot fringes. The other ones have only small formating changes, right?

Thanks Emerson for the review. And you are correct. The other two notebooks are just very minor formatting changes.

tomflexcompute:
> Looks good and also properly implemented!
> 
> When you do these updates I assume you want it to automatically trigger a new docs build? Currently you have to update the submodule state on pre/2.6 in the main repo. However, I can write an action to trigger this automatically if you need.

Thanks Dario. In the previous tidy3d-docs repo, only the change in the latest branch triggers a doc rebuild automatically. I suppose we can do the same here so only the latest docs can change with updates and all previous versions always stay how they are.

PR #29, opened 2024-01-25
Cleaning up adjoint mode converter notebook

momchil-flex:
There were a number of things left over from the initial box handling. We may even want to modify the title image eventually. I cleaned all that up and simplified a few other things. 
- We just define a resolution for the design region, instead of nx and ny number of pixels in the design region, which I believe were too coarse.
- Outside of the design region, we use auto grid. In the design region, we put a mesh override structure the desired uniform resolution. This is how the grating coupler is also done and makes the most sense (no interpolation of custom medium permittivity to simulation grid)
- Just use our in-built binarization filter
- Seems like no need for the conic filter - binarization - conic filter - binarization repetition.
- Also seems like no need for the penalty which again seems to unnecessarily complicate things in the end. We *could* use that penalty somewhere else if it actually improves things.

In the end I got a final design that both has a higher efficiency and looks much better. I think it looks better just because `nx` and `ny` before were too coarse. I am not sure what exactly helped the objective function: maybe it's the same thing, or maybe e.g. the double filtering is not a good idea?

I see some of those things were also carried over to the grating coupler notebook, maybe some day would be good to simplify that one too.

Edit: removed the comment about notebook links, they were fine, I just started off from the notebook in our online python env, which has the links expanded.


AdjointPlugin3InverseDesign.ipynb
tylerflex, line 1: It looks like `dl_design_region` is the resolution of pixels in the design region AND the simulation grid resolution in design region, correct? maybe this can be clarified with a comment in the notebook.  

Also, what's the idea behind setting these to be the same value? is it accurate? I would have assumed we might want a grid size a bit smaller than the pixel size by a few times? should we set a parameter to control this? 

One downside of this high resolution pixel approach is slower post processing right? is it noticeably slower?

tylerflex, line 1: I found it more convenient to set a final beta and then let the increment be determined by the number of steps. The optimization is highly sensitive to beta though, so maybe there's some merit to doing it with a set increment each step?

tylerflex, line 1: By getting rid of the penalty, the only knob we have to tune feature size is the filter radius? I feel the feature size of the new device is a bit smaller than the original, which could explain the improved performance. Maybe it's worth running again with a larger radius to see?

tylerflex (commented):
looks good, thanks for cleaning this up. I think it's much improved and it's interesting that you were able to get good results without the double filter/project and without the penalty.

I think the beta scheduling might have something to do with it (two projections means the beta should be kept pretty low at the beginning) and also the new device seems to have larger features so I dont know if this approach can as reliably get rid of small features. 

So it might be worth seeing what we could get with either:
1. larger filter radius (note: won't really take into account small gaps between features).
2. adding penalty back in, perhaps with a constant `beta=30` or so in the erosion / dilation operations.

AdjointPlugin3InverseDesign.ipynb
momchil-flex: To me it makes the most sense if they are the same, excluding considerations of post-processing time. If the step size in the parameters grid is larger, you're not using all your degrees of freedom (and might not result in a very good structure, I think some of the crudeness in the original final design may be due to that).

The post-processing time in this particular case seems acceptable, it's just 5s.

Regarding an extra parameter - maybe, but I don't think it's needed. Think about how people usually do inv des, like what you would do in ceviche. You just work with the eps array directly. I think no point complicating things here?

momchil-flex: The problem with this for me was that if I wanted to run for fewer steps just to see if things are going well, I would get larger beta jumps (final beta is the same). Now instead, you could run for 10 steps, and then even *continue* for 10 more if you want, with the same beta scheduling as if you ran for 20 from the start.

momchil-flex: I brought the penalty back in, in a separate commit. We can compare the result with the penalty and without, as well as the original sim. I am not sure why you say that the feature size was smaller than the original - there are some pretty small features in the original too. I also wouldn't really say that the new device with the penalty looks better. It does still achieve 97% though!

I think it would be nice to have a notebook that discusses filtering, projection, and penalties in more detail, like showing the effect of a filter with different radius on some starting parameters, and of projection, and plotting the penalty over a given design, somethign which we do in the level set notebook. I don't know if this should be done here, or in a separate example to keep things more atomic?

momchil-flex: Oh actually it looks like the user-side post processing of the gradient is also a bit slower now, it still seems to take a few seconds but it's notcieable.

tylerflex: makes sense

tylerflex: I wonder if it makes sense to have both a beta max and beta increment. just stop incrementing if it gets above max?

tylerflex: maybe a separate example would be best.

tylerflex: I agree with this generally, however I dont think the pixel size ultimately determines the number of degrees of freedom at the higher end, rather I bet it is limited by the number of wavelengths (in material) one can fit in the design region. (or something like this). Basically I bet that doing a grid resolution is overkill but who knows. and if it doesn't hurt us too bad in processing time it doesn't matter too much

momchil-flex: Brought back the penalty properly. I think the device does look better, most of the small features now are around the input/output waveguides because we don't handle that correctly. Efficiency is still 96%!

momchil-flex: Yeah, that's a good thing to consider, but I think there's a finer point here. The maximum number of actually available degrees of freedom, if you could write the most efficient parametrization, is probably indeed much smaller than the number of pixels. However, we are not using the most efficient parametrization - we are using pixels, and in that sense going for lower resolution that the simulation one reduces the capability of our representation to capture the actually available degrees of freedom. Does that make sense?

tylerflex: nice! I think @e-g-melo recently did some treatment to get rid of those btw.

tylerflex: I think I kind of see what you're getting at. As a thought experiment, start from 1 pixel in the design region and increase the number of pixels to infinity while recording the "best" objective function value you get with that number of pixels. At some point, I think we'll see the objective function level off (adding more pixels doesn't help). I guess the question is at what point does that occur relative to the wavelength in the material or the simulation grid size? my guess is the wavelength in the material is probably the quantity that matters here.

momchil-flex:
@tylerflex anything else to do before merging?

tylerflex:
I'm away from computer for a while and can't see the notebook from GitHub mobile. So this last version keeps the penalty in? Can you send the final device as an image in this thread?

momchil-flex:
Yes, it keeps the penalty in. You can review whenever you can no rush! I have a separate version of this for the course.

image

tylerflex:
Looks great! we can merge

PR #30, opened 2024-01-26
Fixed numpy depreciation warnings

tomflexcompute:
There have been a few numpy depreciation warnings about converting an array of one element to a scaler. I found three notebooks that have this issue and fixed them in this PR just to be future-proof. 

I suspect there are more in other notebooks so if you also see any let us know @momchil-flex @tylerflex .

Also from the difference I noticed the waveguide plugin is using a smaller mode plane size now than before, which is good, but whenever we rerun notebooks before release, it might be good to merge the rerun to keep all these details updated.


daquinteroflex:
So having a bit of trouble with the diffs, we're investigating a notebook issue as per the other issue and also when I opened a notebooks PR earlier today, that's why I haven't fully approved till we identify the weird stuff being added.

tomflexcompute:
> So having a bit of trouble with the diffs, we're investigating a notebook issue as per the other issue and also when I opened a notebooks PR earlier today, that's why I haven't fully approved till we identify the weird signs being added.

Oh no worries at all! By no means it's urgent to merge this. Feel free to take your time. In fact I might identify more notebooks with the same issue in the meanwhile and fix them in the same PR...

tomflexcompute:
FYI also lumped in this small fix for https://github.com/flexcompute/tidy3d/issues/1409

tylerflex:
Great thanks, please remember to close that issue once this gets merged.

tylerflex:
I think this is ok to merge?  seems the diff is normal now?
image

daquinteroflex:
Yeah looks good.

 I'm still unclear about the diffs situation I dealt with locally but maybe it was a local thing. Will look into it later.

daquinteroflex (approved):

tomflexcompute:
Thanks both!

PR #31, opened 2024-01-26
handle edge case in erosion / dilation penalty when all(diff==0.0)

tylerflex:


momchil-flex (approved):
This is fine, but shouldn't we have this as an in built penalty if we use it so much?

Note: not sure why but two of the notebooks are listed as having 1000+ lines changed...

tylerflex:
That's really weird. @daquinteroflex do you have any idea what's going on here with the extra lines in the commit? I tried checking out the pre/2.6 versions of the notebooks (Adjoint 12 and Adjoint 6) and adding my edits, but they still have tons of extra additions?

tylerflex:
I will make a separate commit to add the erosion / dilation as a penalty function. I was actually surprised myself that it made it into so many notebooks.

daquinteroflex:
So just to update, I think it's a generalised problem, in a test PR I opened today they appeared too. The weird thing is that the gitconfig is the same as before. A bit tricky to identify exactly what's the deal.

It could have to do with the black configuration?

EDIT: I think nbdime might not be working properly to some level.

tylerflex:
Note, closing as the changes in #48 will take care of it (and this special case is also handled in that penalty class)

PR #38, opened 2024-02-07
GDS export in adjoint and GDS export tutorial #1412.

### LINKED ISSUE #1412: Add examples of exporting adjoint designs `to_gds`.
Users want to export their adjoint designs to gds, but dont know how. In most or all of the adjoint notebooks, we should end with an export to GDS file.

e-g-melo:
Inclusion of GDS export in adjoint notebooks.
Creation of a new Python tutorial on GDS export.
#1412


AdjointPlugin12LightExtractor.ipynb
tylerflex, line 1: looks like the spectrum (Purcell factor) shifted, is that expected?

GDSExport.ipynb
tylerflex, line 1: I think @lucas-flexcompute is working on something similar, just pinging him to make sure there's not too much overlap (although I dont think it would hurt to have several examples of GDS import and export, maybe they could be consolidated or organized into their own sections eventually?)

AdjointPlugin10YBranchLevelSet.ipynb
tylerflex, line 1: This is a pretty minor comment, in the sections

> In addition to a file name, it is necessary to set a cross-sectional plane (`z = 0` in this case) where to obtain the geometry, a `frequency` to evaluate the permittivity, and assign a `permittivity_threshold` to define the shape boundaries in custom mediums.

I would change it slightly to

> In addition to a file name, it is necessary to set a cross-sectional plane (`z = 0` in this case) on which to evaluate the, a `frequency` to evaluate the permittivity, and a `permittivity_threshold` to define the shape boundaries in custom mediums.

This paragraph occurs in a few notebooks.

GDSExport.ipynb
tylerflex, line 1: There are a few other GDS export functions `to_gdstk`, `to_gdspy` I think. It would be nice if we could include each of these in the tutorial, maybe at the end, to showcase all of the various options. Otherwise, the example looks really good!

tylerflex (changes requested):
Thanks a lot Emerson! Just a few minor comments. At a high level, we should figure out
1. coordinate with @lucas-flexcompute , because maybe after this we dont need Lucas to do a separate PR for https://github.com/flexcompute/tidy3d/issues/1411
2. Let's add the other `to_gds*` functions into the GDSExport notebook, and consider adding the `from_gds*` class methods into the GDSImport notebook as well just for completeness? 
3. Some minor grammatical change to the adjoint export paragraph, it also might be worth adding some mention of the other `to_gds` functions there, or a link to the `GDSExport` notebook so that users can know where to go for more details.

Thanks a lot, I think this will really improve our adjoint notebooks and overall documentation!

GDSExport.ipynb
tylerflex: or at the least, we can add the new gds importing functions to GDSImport.iypnb

lucas-flexcompute: Thanks for letting me know. Let's try to merge both versions, because I don't think we should have 2 separate notebooks for the same thing (export GDS).

tylerflex: perhaps we can merge import GDS and export GDS notebooks too? Into a single GDS notebook.

lucas-flexcompute: I think having them separate makes it easier to find and simpler to read.

It's unusual to need both import and export in the same workflow because we not a layout editor (import - edit - export). Either we want to simulate something that is designed outside and available as a GDS, or we design within Tidy3D and want to export the layout. Doing both is not the most common use case. Just my opinion, though.

tomflexcompute: I also think having two separate tutorials, one focusing on import and one focusing on export, might make sense. In general for tutorials I feel like shorter ones are better than longer ones as users can quickly see what they are looking for without having to read over a long notebook.

tylerflex: sounds good to me!

AdjointPlugin12LightExtractor.ipynb
e-g-melo: I've not perceived the shift, but the Purcell factor is sensitive to a slight change in discretization.

GDSExport.ipynb
e-g-melo: @lucas-flexcompute sent me his notebook, and we will merge them. In the end, we will have a much more complete tutorial. Thanks @lucas-flexcompute!

AdjointPlugin10YBranchLevelSet.ipynb
e-g-melo: Thanks @tylerflex! I will make the changes.

GDSExport.ipynb
e-g-melo: After including Lucas's notebook, we will have all of them.

e-g-melo:
Thank you all for the comments! I will make all the changes.

e-g-melo:
Hi! Notebooks were merged and reviewed. Thanks @lucas-flexcompute , @tylerflex , and @tomflexcompute!

AdjointPlugin3InverseDesign.ipynb
lucas-flexcompute, line 1: In the mode field profiles, I usually use the imaginary part for the longitudinal component, because that's the meaningful one. I also like to put all plots of a same mode in the same scale, to properly compare the components. My suggestion would be to use something like (I've also reduced the figure size because that's the maximal size on my monitor):
```python
fig, axs = plt.subplots(num_modes, 3, figsize=(12, 12), tight_layout=True)
for mode_index in range(num_modes):
    vmax = 1.1 * max(abs(modes.field_components[n].sel(mode_index=mode_index)).max() for n in ("Ex", "Ey", "Ez"))
    for field_name, ax in zip(("Ex", "Ey", "Ez"), axs[mode_index]):
        field = modes.field_components[field_name].sel(mode_index=mode_index)
        field.real.plot(label="Real", ax=ax)
        field.imag.plot(ls="--", label="Imag", ax=ax)
        ax.set_title(f'index={mode_index}, {field_name}')
        ax.set_ylim(-vmax, vmax)

axs[0, 0].legend()

print("Effective index of computed modes: ", np.array(modes.n_eff))
```

AdjointPlugin9WDM.ipynb
lucas-flexcompute, line 1: Same comment for the field profile plots as before.

lucas-flexcompute (approved):

AdjointPlugin9WDM.ipynb
e-g-melo: Thanks @lucas-flexcompute! I will take care of changing those plots.

e-g-melo: Updated. 
image

tylerflex (approved):

tomflexcompute (approved):

tylerflex:
should we merge this?

e-g-melo:
Yes... thanks!

PR #40, opened 2024-02-07
Add geometry transformation notebook

lucas-flexcompute:
This is in in regards to https://github.com/flexcompute/tidy3d/issues/1410

@e-g-melo and @tomflexcompute do you have any customer-inspired examples we could use here? Currently this notebook feels very abstract to me. I'm guessing we could use more real-world inspired examples, but the only one we have for sure is the metalens, which already uses transformations.


e-g-melo:
What about something like a [woodpile photonic crystal](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.90.115140)?
image

tomflexcompute:
Thanks Lucas. I think you also need to add the notebook name to the rst file, which in this case would be `structures.rst` I suppose, so that it will be displayed on the website.

lucas-flexcompute:
> What about something like a [woodpile photonic crystal](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.90.115140)? 

That's a great example! Thanks!

e-g-melo (approved):
Thanks for this tutorial @lucas-flexcompute! It is very nice to see how simple is to create complex patterns like those grating couplers.

I found only a few typos:
- Tidy 3D has a rich set -> Tidy3D has a rich set 
- complex simmulations -> complex simulations
- with the transfomration -> with the transformation

Could you include a link to `Transformed` in the introduction?

lucas-flexcompute:
Done, although `Transformed` was not in the docs. I've just created a PR: https://github.com/flexcompute/tidy3d/pull/1458

tomflexcompute (approved):
Just added metadata to the notebook. Thanks @lucas-flexcompute and feel free to merge.

PR #41, opened 2024-02-12
Added plasmonic nanorod array notebook and associated files

tomflexcompute:
There has been some confusion among users regarding how to model a hexagonal (triangular) lattice in FDTD. This example aims to demonstrate that. In addition, relative field enhancement calculation is also a quite common question, which this notebook also demonstrates.


momchil-flex:
This is a great addition for sure, indeed hexagonal lattice is common and we have not discussed it so far!

I understand what you're doing but I disagree a little with this approach
image

I think what we should teach people is that the unit cell is unit cell B. Unit cell is usually be definition something that's periodic. You can still reduce the simulation complexity to that of unit cell A if you apply symmetries in the simulation with unit cell B. In fact that's exactly what you are doing but instead through PEC/PMC boundary conditions. This too is less standard and I think the better way is just to discuss things in terms of the symmetries directly. Furthermore, if the simulation is set up with unit cell B from the start and one wants to try angled incidence, e.g. to compute the band structure as in our other example, they just need to change to bloch boundaries and remove symmetries. What do you think?

`The default automatic shutoff level is 1e-5 but we decreased it to 1e-7 to further enhance the simulation accuracy, which is typically not required.` I think we can add that it can be needed in simulations like plasmonics where there's strong field concentration in relatively small volumes.

tomflexcompute:
Yeah thanks @momchil-flex . That's a good point. The revised version has reflected these changes.

e-g-melo (approved):
Hi @tomflexcompute! Thanks for this nice and helpful example.

I agree with @momchil-flex comment about the unit cell definition in simulation. In addition, I have only this minor typo to report:
- "determine the resonance wavelengths(frequencies)" -> "determine the resonance wavelengths (frequencies)"

tomflexcompute:
Thanks guys. All fixed. If there is no additional comments, I'll merge it later today.

PR #46, opened 2024-02-13
adds `ErosionDilationPenalty` usage to notebooks

tylerflex:
See https://github.com/flexcompute/tidy3d/pull/1467

Note: I only re-ran Adjoint3 (mode converter). The results are fine but slightly different because of some different settings for the erosion beta (fixed after running). There are a lot of warnings, maybe we just re-run them all before 2.6?


momchil-flex:
It seems like the branch has conflicts. Seems like you didn't branch off from latest pre/2.6 here.

We could rerun again later. I just usually don't make a commit where I update *all* notebooks, only some that need updating, so we should keep that in mind.

tylerflex:
> It seems like the branch has conflicts. Seems like you didn't branch off from latest pre/2.6 here.

I think what happened is that Emerson's PR got merged after I made this.

https://github.com/flexcompute/tidy3d-notebooks/commits/pre/2.6/

I can fix the conflicts tomorrow.

tylerflex:
Just implemented this in a fresh branch #48 . closing

PR #47, opened 2024-02-14
Added plasmonic nanorod array notebook and associated files

tomflexcompute:
Sorry I clicked the wrong button and accidentally closed the PR... 

If there is no additional comments, I'll merge it later today.


momchil-flex:
There's a "reopen pull request" button you can click when that happens. :)

Maybe I wasn't too clear before. I think not just the explanation of the unit cell should change, but the simulation definition too. The simulation size in xy should be as shown in the unit cell schematic, and instead of

```
    boundary_spec=td.BoundarySpec(
        x=td.Boundary.pmc(), y=td.Boundary.pec(), z=td.Boundary.pml()
    ),  # pmc on x, pec on y, and pml on z
```

you should use something like

```
    boundary_spec=td.BoundarySpec(
        x=td.Boundary.periodic(), y=td.Boundary.periodic(), z=td.Boundary.pml()
    ),
   symmetry=(1, -1, 0)
```

Doesn't that make more sense?

One thing that should be discussed if we were computing the photonic bands is that since the rectangular unit cell is not the *primitive* one, the bands will be folded. However, since we're not showing bands here, I would say it will only be confusing to mention.

tomflexcompute:
Yeah that's a good point. The reason why I kept the simulation domain that way is because in the paper that's the simulation domain they showed. Also only need to make two cylinders instead of five in this way and the additional three would be just for show and not useful.

momchil-flex (approved):
Thanks, Tom. I think like I said the main advantage now is that one can do angled incidence with a simple modification. But also it's just set up in a more standard way where physical symmetries are applied as simulation symmetries instead of as boundary conditions.

Just one final thing, there's still this comment in the simulation boundary spec definition `# pmc on x, pec on y, and pml on z`.

tomflexcompute:
> Thanks, Tom. I think like I said the main advantage now is that one can do angled incidence with a simple modification. But also it's just set up in a more standard way where physical symmetries are applied as simulation symmetries instead of as boundary conditions.

Yeah indeed as a more general tutorial, it's a good idea to set it up this way.

> 
> Just one final thing, there's still this comment in the simulation boundary spec definition `# pmc on x, pec on y, and pml on z`.

Oh good catch. Thanks! Will fix it and then merge it in a few minutes.

PR #48, opened 2024-02-15
adds `ErosionDilationPenalty` to adjoint notebooks

tylerflex:
Note: #46 had merge conflicts that couldn't be easily resolved so I just re-did the changes on a fresh branch off of pre/2.6.

### DIFF
--- a/AdjointPlugin11CircuitMZI.ipynb
+++ b/AdjointPlugin11CircuitMZI.ipynb
@@ -242,8 +242,8 @@
     "\n",
     "radius = .120\n",
     "beta = 50\n",
-    "\n",
-    "conic_filter = ConicFilter(radius=radius, design_region_dl=float(lx) / nx)\n",
+    "design_region_dl = float(lx) / nx\n",
+    "conic_filter = ConicFilter(radius=radius, design_region_dl=design_region_dl)\n",
     "\n",
     "def tanh_projection(x, beta, eta=0.5):\n",
     "    tanhbn = jnp.tanh(beta * eta)\n",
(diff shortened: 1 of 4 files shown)


tylerflex:
@daquinteroflex the same issue happened here with all of the extra "ghost" lines in the commits. Any idea what is causing this? maybe it has to do with my local jupyter lab? It might be worth someone else checking out this branch , running or saving the notebooks, and seeing if this has any effect.

daquinteroflex:
Sounds good will have a proper test once I've done my final checks on the poetry backend!

daquinteroflex:
I am not completely sure what's going on but I checked out this branch. Added single line space on some notebooks so that the ipynb state would be edited, ran one notebook (but got an error I have not been able to look at) jut to check the diffs. 

https://github.com/flexcompute/tidy3d-notebooks/pull/53

I'm going to try some more reproducible tests to compare ipynb notebook environments (but maybe we might have to live with the spaces for the release or until I can focus on this)

daquinteroflex:
Just in case for you to try tomorrow, to be sure (and narrow down the list of possible causes), does running this make any difference:

```
nbdime config-git --enable --global
```

Otherwise wondering if it might be the ipynb git treatment between different platforms maybe? It's just a bit hard for me to reproduce.

tylerflex:
> nbdime config-git --enable --global

That seemed to not do anything, at least after I ran it, there were no changes to the notebooks. Should I run a command after this perhaps?

tylerflex:
Related, are there some internal docs on how to use `nbdime`? or whatever other tool we use for cleaning notebook conflicts / removing metadata?

daquinteroflex:
I am adding docs on this in the next updates to the `develop` cli PR but likely for 2.7. I was thinking it should have already been automatically installed by:

https://github.com/flexcompute/tidy3d/blob/pre/2.6/.gitconfig

per the instructions that apply in our case from the intiial tidy3d-docs instructions:

https://github.com/flexcompute-readthedocs/tidy3d-docs?tab=readme-ov-file#setup

It's weird as I think it should just sort it already inherently when running git commands, maybe try if you rebase if it does anything different?  Possibly, one idea is to add poetry in the .gitconfig to guarantee the commands of nbdime should be run within the correct environment where their cli is available. I am wondering if you were to copy, rename and add the files again if they'd look any differnt?

tylerflex:
ok I will play around with this today and see if I can resolve it. It seems only the WDM notebook is affected so worst case I can just re-do that one and see if it works.

tylerflex:
Seemed to have mostly fixed it (was almost 100k lines changed now a few k) still not sure exactly why there are that many. but maybe this is ok to go into pre/2.7 after the release.

momchil-flex (approved):

PR #51, opened 2024-02-15
added adjoint optimization of phase mask

tylerflex:
image
A couple open questions:

1. The penalty really degrades performance (the logo looks a bit rough and not even). I left the penalty function in case users want to play with it but set the strength to 0.

2. I originally used a range (0.1, 0.9) within which to evaluate the intensity. So if the intensity values were above 0.9, 
they were just considered as 0.9, as a way to kind of encourage smoother structures. I'm not sure exactly how well this ended up working, so I set the range to (0,1) but again left it in case the user wants to play with it.

I could be persuaded to remove both 1 and 2 if it's thought that these just complicate things unnecessarily.


e-g-melo (approved):
Thanks @tylerflex! The final image is very beautiful!

I have only a few comments:

1 - It would be nice to have a comment on why you chose `thickness = 4 * wavelength / 2 / np.pi / (np.sqrt(permittivity) - 1)`.
2 - In [5]: "side for visualziation" -> "side for visualization"
3 - Include space before comment in "penalty_loss = 0#penalty_weight * penalty_fn(params, beta)".
4 - Why is `penalty` always zero throughout the optimization? Is that expected?

Thanks!

tylerflex:
Thanks Emerson, I'll make the changes tomorrow.
As for the penalty, I explain a bit above

> The penalty really degrades performance (the logo looks a bit rough and not even). I left the penalty function in case users want to play with it but set the strength to 0.

The result is seemingly more impressive without it. So I'm not sure if it just makes it confusing to include.

tylerflex:
I made the penalty switching more built in (with a comment too) and included this in the objective
```py
    if include_penalty:
        penalty_loss = penalty_weight * penalty_fn(params, beta)
    else:
        penalty_loss = 0.0

```

e-g-melo:
Got it! I think that should be fine. The users will understand and turn the penalty on if they wish.

momchil-flex:
Nice! Some small comments:
 
```
# resolution
min_steps_per_wvl = 24 / np.sqrt(permittivity)
```
The idea of `min_steps_per_wvl` is to not have to define things in terms of permittivity. I am not sure what is the logic here?

- `We'll design this phase mask to give an transmitted` -> "a transmitted"

- Ignoring the exact shape of the transmitted intensity, this looks kinda like a metalens that is trying to focus light at a very close focus. Have you tried what happens if you put the monitor further away?

tylerflex:
> # resolution
> min_steps_per_wvl = 24 / np.sqrt(permittivity)
> The idea of `min_steps_per_wvl` is to not have to define things in terms of permittivity. I am not sure what is the logic here?

The `JaxCustomMedium` has a validator on the maximum number of cells. I originally had set the resolution like this so that for a given side length, I could get the same number of cells as I changed the permittivity. Although I think we can remove it.

> * `We'll design this phase mask to give an transmitted` -> "a transmitted"

Fixed

> * Ignoring the exact shape of the transmitted intensity, this looks kinda like a metalens that is trying to focus light at a very close focus. Have you tried what happens if you put the monitor further away?

Yea it seems very dependent on the distance from the metalens. I've tried many combinations. I think the relevant quantities are the feature size of the logo, the focal length, and the mask diameter, all relative to the wavelength. The parameters I chose here seemed to do the best job at balancing all three of these. As we make the focal length larger, I think the performance would be better IF we could increase the diameter, but there's this number of pixel limitation so I had to stop somewhere.

Would be good to relieve those limitations a bit, which we have planned in https://github.com/flexcompute/tidy3d/issues/1111 but need to find a solid chunk of time to work on that. Initial testing showed that it was generally limited by the number of calls to `xarray.interp` so maybe we can find some way to improve that, or better approximate the integrals.

PR #54, opened 2024-02-16
Add PreFab-enhanced adjoint grating coupler optimization notebook

tylerflex:


tylerflex:
@Dusandinho Thanks for this notebook. I think it looks really good, basically ready to publish already but I wanted to add a few suggestions and then get @e-g-melo  to take a look at it when he has time as he's got a lot of inverse design expertise. A few things to start:

1. The introduction of pre-fab happens pretty late in the notebook. Can you think of any way we can dramatically shorten the Tidy3D bits copied from the grating coupler notebook or is it unavoidable?  For example, in addition to loading the grating coupler design from file, maybe we can load the functions and variables we need to generate it as well? Perhaps using `pickle` or `dill`? That way we can focus more on prefab earlier on.
2. I like the way you do installation and authentication in the notebook, might have to borrow some of those ideas for Tidy3D some day :).
3. Do you want to add any more diagrams or links to introduce prefab? For example in the introductory paragraph? It might be good for users clicking on this example to learn right off the bat what prefab (generally) is doing. I do feel like a diagram / image and some links might be helpful.
4. I feel like some more explanation of what is a "prediction", "Correction", and "outcome" could be useful around cell [29]. Again maybe an image (btw these images can be put in the `img/` directory of the repository.
5. After cell [30] what exactly do the pink contour lines correspond to? We're taking the predicted / corrected / outcome images and then binarizing them, which gives the pink lines?
6. After cell [33], perhaps it would nice to plot these loss values. For example, plot loss vs binarization % with one line for prediction and one line for corrected outcome. I think this will drive the point home more, in addition to the text output.
7. It could be useful to describe how to export the final design to GDS file as most of our users ask us how we can do that after our adjoint runs. Is this something you can do through prefab directly (sorry if I missed it).

No rush, but I think if we can make these changes it make the example even better! I mostly want to make sure people reading it will understand the power of your tool and it doesn't get lost.

@e-g-melo  maybe after this could you take a look to see if you have any technical comments?

Thanks!

e-g-melo:
Hi @Dusandinho and @tylerflex! Thanks for sharing this notebook. The tool is very interesting for sure. 

I can take a look at it and also help to include some metadata we need to make the notebook available in our docs properly.

For now, I have one question: Can we use a GDS file directly instead of starting from a `numpy` array?

Dusandinho:
@tylerflex Thanks for the suggestions. I definitely agree that they will improve the notebook. I'll make the changes shortly.

@e-g-melo @tylerflex I wasn't sure if it made the most sense to start from optimization again or to just import and start from the final design. I'm open to either. For example, we can load the design in from GDS, run the PreFab predictions/corrections, run the simulations, analyze the results, then export the correction back to GDS.

e-g-melo:
Starting from the final design seems more straightforward and general to me. So, users would easily understand that they can apply Prefab to any existing design.

Dusandinho:
@tylerflex I made the suggested changes. Please let me know the best way to add it with this new pr.

tylerflex:
Hm perhaps for now you can just email me the notebook. Unless you are able to figure out how to push a commit to this branch. Sorry to complicate things! See you tomorrow

tylerflex:
@Dusandinho I incorporated your changes in [this commit](https://github.com/flexcompute/tidy3d-notebooks/pull/54/commits/0567285575d4c79214a89234d0570d569d7686eb) and made a few edits again so it would get compiled properly.

I think this new version is very nice! 

I think the only potential improvement I can think of is whether we can somehow load the "base" simulation to file, and then we won't need all the global variables and the `run_simulation(device)` function can simply load it from file and add the device before running it. Unless you think the global variables (including all of the components, like the sources and monitors) are nice to have defined? But yea overall I think it's much better now!

Maybe I can take a stab at this ^ since it doesn't involve pre-fab.

tylerflex:
@Dusandinho the current notebook seems to write the corrected GDS file into the same file that was used as the original? Does this mean that if the user runs the notebook twice, they will load the corrected file on the 2nd time? I'm considering changing it so we write the final device into a new gds file, unless there's some subtlety I'm not understanding.

tylerflex:
Oh, so maybe it's being written into a different cell? That's probably ok then?

Dusandinho:
@tylerflex Great! Yes, please feel free to make changes to the simulation setup.

The PreFab results are written to a different cell in the same file, with the different variations overlapped so the user can compare the differences (the layer numbers are a bit arbitrary though). The only problem I see would be if they export twice, there will be an error due to having the same cell names. We can definitely just export to a different file if you see fit.

Dusandinho:
Though it just occurred to me that the naming issue would still be there when writing to a different file 🤦‍♂️

I can restructure the export code a bit.

tylerflex:
I see, so if I run the last cell twice it might error? I can test.

I'm doing quite a bit of updates, I'll push in a bit:

1. Saved a "base" simulation (no device) to file, removed all of the simulation variables.
2. Made a `make_simulation(device)` to just construct the `gc` from `device` and add this to a copy of base.
3. Refactored `run_simulation` to use this.
4. set global variables for the gds file and cell names.
5. will make it so that if the user re-runs the last cell, it will simply overwrite the previous values in the cells.
6. will run the sweep in a `Batch` (in parallel) so it will be much faster.

tylerflex:
> I can restructure the export code a bit.

Maybe wait until I push my new notebook, unless you want to work only on that last cell. Otherwise I can fix the merge conflicts later, no worries.

tylerflex:
It seemed to not error when I re-ran the last cell. Will push my changes after the meeting for you to take a look at. Feel free to edit anything that seems not quite right.

tylerflex:
Ok I'm done with my edits, you can check them out in this file. Let me know if you have edits, or feel free to download and edit directly.

https://github.com/flexcompute/tidy3d-notebooks/blob/6cd5aacc2a9fe5dfd6e66780f4997f223a05287f/AdjointPlugin14PreFab.ipynb

Dusandinho:
The edits all look good to me, thanks.

Batch simulation is going to be very useful for this variability testing.

tylerflex:
Sounds good.

@e-g-melo 
Could you take another look and see if you have any other comments? Then we can wrap this up.

e-g-melo (approved):
Hi @Dusandinho and @tylerflex! You have done an impressive work on this notebook! Thanks for all these improvements!

I have only two suggestions:
1. I suggest you move the "Grating Coupler Simulation Configuration" session just before [8] and then include the grating coupler GDS image in the sequence, as below (Including the necessary adjust like the session titlle. But feel free to ignore my suggestion.

image

image

image

image

image

2. It was not immediately clear to me what you refer to in [19] regarding "constrained" and "unconstrained". That part could be expanded a bit to give more context to the reader.

Dusandinho:
@e-g-melo Thank you! I'll let @tylerflex comment on the first item. Though I think it makes sense to move.

As for the second, the constrained design is from the previous notebook (with the feature size penalty and larger filter size), and the unconstrained design is the current one (without the penalty). But I now realize we don't actually define these terms anywhere. @tylerflex Can these be added somewhere early on? Maybe there are better names we can use.

tylerflex:
Sure, I'll take care of both of these and then we can publish this notebook shortly. Thanks @e-g-melo

tylerflex:
Ok I did some restructuring. I think it improves things. @e-g-melo @Dusandinho feel free to take another look.

tylerflex:
I also changed the title to "Machine learning-based fabrication constraints for inverse design using PreFab and Tidy3D" to get that ML buzzword in :) 

@tomflexcompute could you take a read through? I think I'm too close to it now to notice if it still reads well.

e-g-melo:
Thanks, @tylerflex and @Dusandinho! I have no other comments here. The notebook is pretty good!

Dusandinho:
Edits and new title look great!

tomflexcompute (approved):
This is very interesting for sure. And congratulations to @Dusandinho for developing such an amazing tool, which can greatly push inverse design to industrial applications. 

When I went over the notebook without any prior knowledge about prefab, the first image is a bit confusing. Initially I wasn't sure what prediction and correction mean and what each image represents. After reading the "Set up PreFab" session, everything becomes very clear and the second image is a very good demonstration of the idea. I wonder if it makes sense to move the current first image to after the current second image as another "real-world" example, instead of having it as the first image when readers haven't had a good idea of what it means. 

A minor question I have is why the notebook file name is "AdjointPlugin14"? The last one we have is "AdjointPlugin12". Is there an "AdjointPlugin13" coming later?

Also a small typo: 
"indiciates" -> "indicates"

tylerflex:
Thanks @tomflexcompute

I'll try moving the images where you said below.

As for the 14, there's also #51 which is 13 as it was created earlier. Waiting on final approval for that but I think it will go in around the same time as this.

Will fix typo

tylerflex:
How about like this, we show the first image and then say "here's another example with a more complex geometry"

image

tomflexcompute:
Yeah sounds very good. Thanks.

I didn't realize there is another PR for adjoint example 13. All are great examples!

Dusandinho:
Wonderful! Thank you @tylerflex, @e-g-melo, and @tomflexcompute for the great input and work. So happy that we were able to get this notebook done together.

tylerflex:
Yea thanks @Dusandinho ! I'll let you know when the notebook is made public, probably tomorrow.

PR #68, opened 2024-02-27
Inclusion of structural color notebooks.

e-g-melo:
Hi! I've included two notebooks related to metamaterial structural colors.

- The MIMResonator only reproduces a MIM structure from a paper.
- The AllDielectricStructuralColor reproduces an existing GUI example. I've used it as a case study to run a parameter sweep using the new Design plugin. The idea is to use it to test the compatibility between Python and the future Design implementation in GUI.


AdjointPlugin9WDM.ipynb
tylerflex, line 1: let's not change this file.

AllDielectricStructuralColor.ipynb
tylerflex, line 1: It will be great to have another design plugin demo, thanks!

* Let's also link to the `Design.ipynb` tutorial notebook in the introduction section.
* In `build_sim` and other functions, I think it's useful to add type annotations, for example `def build_sim(period: float = period) -> td.Simulation` and also a docstring underneath.
* using links to the API reference for the various `tdd.` classes (like `tdd.Method`) I think will be very useful around cell [7].
* [11] the "reflectance" column seems like it has python objects in it? maybe we should just store a list or numpy array unless there's a good reason for storing the full object?
* For the final plot [14], I think it would be useful to specify what the legend values (periods) are, maybe like `period={} um` in the legend. Also including the units. When I first glanced at it, I wasn't 100% clear what the different lines were referring to.

Thanks a lot for this notebook, looks great! and demonstrates the phenomenon very well and the design plugin.

docs/case_studies/metamaterials_gratings_periodic.rst
tylerflex, line 1: I wonder if we should also add to the "design space exploration" section. Is it bad to have duplicate entries for the same notebook? @tomflexcompute do you know? otherwise it's fine like this.

MIMResonator.ipynb
tylerflex, line 1: Notebook looks great, thanks @e-g-melo 

* After cell [2] `Material definition.` -> `Next, we define our materials.` sounds a little more natural here.
* Before cell [5], there's a link to the examples library. I think it would be more useful to link to 1 or 2 specific examples that are similar to this. I can imagine a user might not know where to go after clicking on that link.
* After cell [6] ` if everything are on the correct place` -> `if everything is in the correct place`
* Let's add type annotations and docstrings to the function whenever we can (similar to comment on the other notebook). I don't always remember to do this, but I think it really helps code readability to know what the function accepts, what types it returns, and also a quick description of what the function does.
* Before cell [8]. Could we expand just a little and mention that we use `web.run_async` to run this set of simulations in parallel? This function doesn't come up so often in the notebooks so I think it will be great to introduce it just a little bit with a sentence and link to the API [reference](https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.web.api.asynchronous.run_async.html). Thank you

metadata/metadata.txt
```diff
@@ -112,6 +112,12 @@
     "description": "This notebook demonstrates the integration of PreFab into Tidy3D to incorporate advanced fabrication constraints for photonic inverse design..",
     "keywords": "inverse design, adjoint optimization, prefab, grating coupler, photonics, Tidy3D, FDTD",
     "feature_image": "N/A"
+  },  
+    "AllDielectricStructuralColor.ipynb": {
+    "title": "How to build and simulate an all-dielectric structural color device using Tidy3D | FDTD",
+    "description": "How to build and simulate an all-dielectric structural color device using Tidy3D | FDTD.",
+    "keywords": "structural color, all-dielectric, color filter, design, parameter sweep, Tidy3D, FDTD",
```
tylerflex, line 119: maybe add `design plugin` too. for SEO, in case someone searches "Tidy3d design plugin" on google.

tylerflex (approved):
Thanks @e-g-melo just a few comments but overall looks very good. Consider me "approved" once the comments are addressed!

tylerflex:
Also, I think the structural color notebook could be great for advertising our design plugin through some marketing materials / linked in post. It's a great application of the plugin.

tomflexcompute (approved):
Thanks Emerson for the great additions!

 In the all-dielectric structural colors notebook:
 
 - Can we reduce the size of the thumbnail image just slightly (10-20%)? Now it looks a bit larger compared to the thumbnails in other notebooks. 
 
 - The peak reflectance goes slightly above 1 in some cases. Usually it means the grid is not fine enough. Do you think that's the case here? Having flux data above 1 is a common question from users. Maybe we can use a slightly finer grid to resolve it? 
 
 - "latter" -> "later"
 - "... name argument match the ..." -> "... name argument **matches** the ..."
 - "build-in" -> "built-in"
 
 In the MIM resonator notebook:
 
 - Can we reduce the thumbnail image size slightly as well?
 
 - Can we add a reference to another [MIM example](https://www.flexcompute.com/tidy3d/examples/notebooks/PlasmonicWaveguideCO2Sensor/)?
 
 - One thing I also noticed is that when using auto aspect ratio, the green arrow of the source is often displayed in a strange position. Do we have control over it? @tylerflex 
 
 - "Fist" -> "First"
 - "... submit an run all ..." -> "... submit **and** run all ..."
 - "if everything are on the correct place" -> "if everything **is in** the correct place."
 - "we will look the magnetic field" -> "we will look **at** the magnetic field"
 
The `metadata` folder and the files in it were introduced to systematically check and change metadata for all notebooks. At this point it looks like all systematic SEO efforts have been done so the `metadata` folder can in principle be retired, as long as we still keep adding metadata into the notebooks themselves. Shall we just remove the entire folder?

AdjointPlugin9WDM.ipynb
e-g-melo: Yes.. absolutely!

MIMResonator.ipynb
e-g-melo: Thanks @tylerflex! I will modify the notebook and include the `run_async`.

AllDielectricStructuralColor.ipynb
e-g-melo: I've included the monitor data as an object because it was a solution I found to merge the current GUI sweep functionalities (show 1D and 2D monitor data for each simulation) into the Design plugin. So, to visualize this notebook in GUI, we should have the same structure.

tylerflex: ok I see

e-g-melo:
> Thanks Emerson for the great additions!
> 
> In the all-dielectric structural colors notebook:
> 
> * Can we reduce the size of the thumbnail image just slightly (10-20%)? Now it looks a bit larger compared to the thumbnails in other notebooks.
> * The peak reflectance goes slightly above 1 in some cases. Usually it means the grid is not fine enough. Do you think that's the case here? Having flux data above 1 is a common question from users. Maybe we can use a slightly finer grid to resolve it?
> * "latter" -> "later"
> * "... name argument match the ..." -> "... name argument **matches** the ..."
> * "build-in" -> "built-in"
> 
> In the MIM resonator notebook:
> 
> * Can we reduce the thumbnail image size slightly as well?
> * Can we add a reference to another [MIM example](https://www.flexcompute.com/tidy3d/examples/notebooks/PlasmonicWaveguideCO2Sensor/)?
> * One thing I also noticed is that when using auto aspect ratio, the green arrow of the source is often displayed in a strange position. Do we have control over it? @tylerflex
> * "Fist" -> "First"
> * "... submit an run all ..." -> "... submit **and** run all ..."
> * "if everything are on the correct place" -> "if everything **is in** the correct place."
> * "we will look the magnetic field" -> "we will look **at** the magnetic field"
> 
> The `metadata` folder and the files in it were introduced to systematically check and change metadata for all notebooks. At this point it looks like all systematic SEO efforts have been done so the `metadata` folder can in principle be retired, as long as we still keep adding metadata into the notebooks themselves. Shall we just remove the entire folder?

Yes, I think we should. That can simplify the inclusion of new notebooks a bit.

I will rerun the simulations with a higher resolution to solve the flux > 1. Thanks!

e-g-melo:
Thanks, @tomflexcompute and @tylerflex for the great comments! I will make the changes you have proposed.

e-g-melo:
I've updated the notebooks addressing all the points you have mentioned. Regarding the flux > 1, I increased the grid resolution, but that didn't help. Then, I modified the axis limits to make it less obvious. 

I wonder if we could have an option to make the output less verbosity in the case of even larger parameter spaces.

tomflexcompute:
Thanks @e-g-melo . Not sure if reducing `shutoff` would help to eliminate >1 flux. If not, that's probably fine. I think you can merge it as it is.

PR #70, opened 2024-02-28
fixes import error in adjoint 3

tylerflex:
we should deploy this whenever possible, adjoint notebook 3 will error because I messed up one of the imports.


daquinteroflex (approved):

daquinteroflex:
Just requires updating the latest branch in tidy3d submodule and it should propagate to the docs I believe.

tylerflex:
ok, just pushed, we'll see how it goes!

PR #71, opened 2024-02-29
Adding a new example: thermo-optic modulator with doped silicon heater.

marc-flex:
Not much new with respect to previous modulator examples except that this combines both thermal and free carrier effects simultaneously.


dbochkov-flexcompute:
- Seems like these lines needs to be corrected:
> The scketch below highlights the geometric features of the considered waveguide: ADD FIGURE HERE
>
> We begin by importing the required modules for this example. Notice that a module named tutorial_utilities has been created for this tutorial and includes some convenience functions.
- cell 1: remove 
```
%reload_ext autoreload
%autoreload 2
import tutorial_utilities as tu
import copy
```
- cell 2: remove commented code `#x_i = 0.55` and everywhere else
- cell 10: need to change coefficients so that it consistent with https://doi.org/10.1109/JPHOT.2011.2171930
```
ne_coeff = -1.91e-21
ne_pow = 0.992

nh_coeff = -2.28e-18
nh_pow = 0.841

k_factor = wvl_ref * 1e-4 / 4 / np.pi  # factor for conversion from absorption coefficient into k

ke_coeff = k_factor * 3.22e-20
ke_pow = 1.149

kh_coeff = k_factor * 6.21e-20
kh_pow = 1.119
```
- cell 24: let's make optic simulation smaller by some buffer
```
buffer = 1
...
    size=(heat_sim_width - 2 * buffer, h_cladding+h_box - 2 * buffer, 2.*wvl_um),
```
- cell 25: remove 
```
cs_grid = optic_sim.discretize(box=td.Box.from_bounds(*electrons_data[0].bounds))
```
use `x=target_grid.x, y=target_grid.y` in `interp`s, this and reducing optic sim size previously will eliminate a bunch of warnings
- cell 28: switch to `run_mode(mode_solvers_thermo_electric[n], results_file=name, reduce_simulation=True)` to eliminate some of warnings
- cell 34 (n_eff plot): missing axis labels
- cell 30: move reference line from -1 to 1

marc-flex:
@dbochkov-flexcompute I have addressed your comments.
I'm sorry for some of them: I was sloppy

dbochkov-flexcompute:
Looks good, added some small corrections (changed fig, renamed file, etc)

tomflexcompute:
Congratulations @marc-flex for the first notebook! Great addition to our example library. A few comments:

- Could you add the notebook to the docs/case_studies/pic_active.rst so it can be displayed correctly in the docs and learning center?
- Usually we use a 3D rendering as a schematic to illustrate the idea as well as work as the thumbnail in the learning center. Maybe you could use this image that I just made?
doped_silicon_heater
- In the last plot, the axis label of the second plot overlaps with the first plot. Could you use tight_layout to avoid it?
- I didn't read the referenced paper closely. Is our result consistent with the reference?
- I think the metadata is missing. Here are the [instructions on](https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e) how to add it. If you are not sure about it, I can add it later so please don't worry about it for now. 

A few typos that I found:
'scketch' -> 'sketch'
'wave guide' -> 'waveguide'
'concenrations' -> "concentrations"
'difine' -> 'define'
'therfore' -> 'therefore'
'multiply by' -> 'multiplied by'?
'contact currents have units if A per unit length' -> 'contact currents have units **of** A per unit length'
'embeded' -> 'embedded'
'permitivity' -> 'permittivity'
'Note that since ther permitivity of the material' -> 'Note that since **the** **permittivity** of the material'
'strucure' -> 'structure'
'lenght' -> 'length'
'devide' -> 'divide'
'doted' -> 'dotted'
'termal' -> 'thermal'
'sence' -> 'sense'
'... we can now solve the charge distribution solver for different applied biases' sounds like what you are trying to say is `... we can now solve the charge distributions for different applied biases`

Adding @e-g-melo here for taking another look too.

marc-flex:
@tomflexcompute thanks for reviewing!

The latest commit addresses all your comments. The only thing that'd be missing is the image, both from metadata and the notebook itself. I couldn't download it? If you could send it to me I can include it both to the notebook and the metadata.

With regards to the data in the reference: even though we are somewhat following the example, we can't use their doping values. This is because when we do, some parameter gets out of bounds (I can't remember exactly what it was, @dbochkov-flexcompute might know what it was) and we got a bunch of warnings. Long story short, we can't compare directly. I have, however, changed the x-axis to power so that it can compare qualitatively.

daquinteroflex:
Hi Marc and Daniil, really interesting example and looks really good!

Just a few tiny suggestions:
* Does it make sense to move ` from devsim.umfpack.umfshim import local_solver_callback` to the top of the file? Just since that way it errors earlier.
* A little subsection on cell 8 would help users jump between the tidy3d meshing and devsim meshing

What do you think about this?
* Just in terms of actual physics and what we're representing, in my understanding the thermo-optic coefficient is also a function of temperature even if it is nearly constant at room temperature (which is set to 300K as the reference temperature). This is more important in cryogenic variations as per the figure below, in which case also the charge effects vary, but also interesting in the context of having our active doped area heated up to around 400-500K. In terms of these dynamic "constants" changing as the device is heated, we're assuming that some of these constants don't vary dramatically according to a given heating condition? Isn't this also a coupled function to the applied optical frequency so that we need to just remind the users these values are 2um specific so they don't reuse them for 1550nm. I do think this simulation physics implementation is valid - but maybe just explicitly mentioning this caveat possibly where it mentions so users are aware of this assumption in terms of accuracy:

```
Note that since _ther_ permittivity of the material is a function of wave frequency, we require a reference wavelength. A reference temperature must also be given for the same reasons
```
(note the typo too)

From:
https://pubs.aip.org/aip/apl/article/101/4/041905/111710/Thermo-optic-coefficient-of-silicon-at-1550-nm-and

image

It'd be interesting to work out the expected "contant variation" error, but this could be probably determined by the user once we mention this caveat.

Just another thought, but maybe not for this notebook since it's ready to go. It is quite interesting how we could simplify the definition of constants between devsim and tidy3d in order for, say users to be able to do easier parametric design between relevant design variables. We would have for example used this notebook in our research group and iterated over multiple design variables - maybe using the design plugin, or just manually functionally. I guess it'd just involve putting multiple code blocks in a function, but that's easy for the user to do really.

Possibly it'd also be interesting to have the last plot compare to the experimental results to justify accuracy? (Apologies for possible extra work? Maybe we believe it's already accurate enough?)

image

e-g-melo (approved):
Hi @marc-flex! Thanks, this is a great addition to our examples!

I have only some minor comments.

1 - In the introduction and after [1], you have used both "p++ - p - p++" and "p++/p/p++". I think you should keep only the former throughout the notebook.
2 - It is impossible to observe differences between the hole density plots in [6] (probably very small). Is there a way to highlight them?
3 - I'd suggest to include comments/units in those `xx_coeff` and `xx_pow` parameters in [10].
4 -  "A reference temperature must also be given for the same reasons" -> "A reference temperature must also be given for the same reason."
5 - You could include the names of the auxiliary structures in this sentence: "Notice that we have created an auxiliary silicon structure".
6 - The last chart is "Phase Shift x Power", so I suggest to include more details in this sentence "As it can be seen, we'd need to apply a voltage ~ 3V in order to get a phase shift of pi".
7 - I suggest you include a header after [2], maybe "Charge simulation"
8 - The same after [7]: "Heat simulation".

dbochkov-flexcompute:
> * Just in terms of actual physics and what we're representing, in my understanding the thermo-optic coefficient is also a function of temperature even if it is nearly constant at room temperature (which is set to 300K as the reference temperature). This is more important in cryogenic variations as per the figure below, in which case also the charge effects vary, but also interesting in the context of having our active doped area heated up to around 400-500K. In terms of these dynamic "constants" changing as the device is heated, we're assuming that some of these constants don't vary dramatically according to a given heating condition? Isn't this also a coupled function to the applied optical frequency so that we need to just remind the users these values are 2um specific so they don't reuse them for 1550nm. I do think this simulation physics implementation is valid - but maybe just explicitly mentioning this caveat possibly where it mentions so users are aware of this assumption in terms of accuracy:

True, we should explicitly mention that we assume linear relationship between permittivity and temperature. And also change coefficient to 1.76×10−4 K^-1 since we consider 2um, same as done in the referenced work.

In regard to more exotic cases where temperature perturbations are not well described by a constant thermo optic coefficient: in those cases I would just describe change in permittivity (and other properties) as a general function of temperature using our `CustomHeatPerturbstion` class, and not invoke notion of thermo optic coefficient at all, since it kinda assumes linear relationship

marc-flex:
Hi Emerson,

> 2 - It is impossible to observe differences between the hole density plots in [6] (probably very small). Is there a way to highlight them? 

I tried with different scales but the equilibrium hole distribution does not seem to change significantly between cases. Electron distribution does change a lot, but since we're doping with holes it kind of made sense to plot holes. Pasting the electron distribution for your reference.
image


>3 - I'd suggest to include comments/units in those `xx_coeff` and `xx_pow` parameters in [10]. 

These parameters are described in the reference given in the commented text given just before they're introduced. This does, however, highlight a possible mismatch in units. The units of these coeffs are given in the reference (see image below), which confuses me since I thought these were dimensionless parameters. Worse still, we use these later in `nk_to_eps_sigma()` without any regards for units. Can any of you help me here? @dbochkov-flexcompute @daquinteroflex 
image

daquinteroflex:
Hi Marc, 

So I believe the $\Delta N_e$ and $\Delta N_h$  units are $\text{cm}^{-3}$ per the image below that text (also I think doping concentrations get defined that way commonly I think?), and if $p$ and $r$ are ${cm}^{3}$ then  they cancel out to dimensionless refractive index at least conceptually. But agree there's something weird where they define, 

> and $cm^3$ for $\Delta$ n

I've not encountered a definition like that before for refractive index changes, and am not sure that's right actually within the paper. Maybe there's something I've missed in terms of the implementation though, but definitely curious for me at least.

image

dbochkov-flexcompute:
> These parameters are described in the reference given in the commented text given just before they're introduced. This does, however, highlight a possible mismatch in units. The units of these coeffs are given in the reference (see image below), which confuses me since I thought these were dimensionless parameters. Worse still, we use these later in nk_to_eps_sigma() without any regards for units. Can any of you help me here?

> So I believe the ΔNe and ΔNh units are cm−3 per the image below that text (also I think doping concentrations get defined that way commonly I think?), and if p and r are cm3 then they cancel out to dimensionless refractive index at least conceptually. But agree there's something weird where they define,
> 
> > and cm3 for Δ n
> 
> I've not encountered a definition like that before for refractive index changes, and am not sure that's right actually within the paper. Maybe there's something I've missed in terms of the implementation though, but definitely curious for me at least.

yeah, it looks like there are some inconsistencies in the paper. As of now, units of a, c, p, and r are actually dependent on values of b, d, q, and s. I think probably a better way would have been to present these dependencies as: dn = (dNe/p)^q + (dNh/r)^s. For the purposes of this notebook I think we can just skip units for these parameters. These parameters are not used anywhere outside of these formulas and we know that their outcome has the right units (refraction index and conductivity).


> > 2 - It is impossible to observe differences between the hole density plots in [6] (probably very small). Is there a way to highlight them?
> 
> I tried with different scales but the equilibrium hole distribution does not seem to change significantly between cases. Electron distribution does change a lot, but since we're doping with holes it kind of made sense to plot holes. Pasting the electron distribution for your reference.

what if you just plot the difference in holes and electron concentrations? I guess both could potentially be contributing (even if negligibly little) to permittivity changes
```
holes_diff = holes_data[-1] - holes_data[0]
electrons_diff = electrons_data[-1] - electrons_data[0]
holes_diff.plot(grid=False, ax=ax[0])
electrons_diff.plot(grid=False, ax=ax[1])
```

tomflexcompute (approved):
Thanks @marc-flex and @dbochkov-flexcompute . All look good to me.

tomflexcompute:
Is the PR ready to be merged or are we waiting on something?

daquinteroflex:
Looks good to me too to merge if you want! I can look more deeply into the
verification which I was curious about in my feedback at some point in the
future. It’s great already!

marc-flex:
@tomflexcompute I forgot about the last of Daniil's suggestions. Let me adress that and commit. Also, I can squash commits if that's preferable.

marc-flex:
@tomflexcompute  I have addressed all comments and squash all commits (you'll see the last force-push).

dbochkov-flexcompute (approved):

daquinteroflex (approved):

marc-flex:
Should I press the green button? (i.e., `Merge pull request`)

tomflexcompute:
Thanks @marc-flex . Looks good to me.

PR #72, opened 2024-02-29
add data file for prefab notebook and make exception in gitignore

tylerflex:


daquinteroflex (approved):
Yeah that makes more sense

PR #78, opened 2024-03-06
Clarify the transmission definition

tomflexcompute:
- Clarify the transmission definition in the Y branch notebook. 

- Modify the grid resolution in strip-to-slot converter so flux<1.

@alec-flexcompute feel free to point out anything else related to this PR.


alec-flexcompute:
There's another instance in the 1x4 MMI power splitter

alec-flexcompute:
There's also an instance in "Parametrized level-set optimization of a Y-branch"

alec-flexcompute:
Otherwise I think that's all cases. Thanks Tom!

tomflexcompute:
Ok thanks. The adjoint notebook is a bit hard to rerun and there is also a comment in the markdown about why there is a multiply of 2 so I'll leave that as it is for now.

PR #81, opened 2024-03-07
Added WGA example notebook

tomflexcompute:
Hi @prashkh just want to have you review this PR. I figure it could be a good way to get familiarized with the typical workflow of contributing a new example. Also would really appreciate your review. @e-g-melo and I have been doing it routinely so essentially add the notebook, add a thumbnail image as well as any schematics, and add the notebook to the rst file.


prashkh (approved):
This looks great! Nice job on the simulation which matches with the paper quite well. My small comment would be for the polar plots to normalize to [0,1] range for the intensity value. This way the directionality is clear.

e-g-melo (approved):
Hi @tomflexcompute! This is a very exciting and didatical example! Thanks! 

I have only one suggestion:
Rephrase this sentence "we defined the central frequency of the pulse to be that at 1550 nm and the frequency width to be 10% of the central frequency." to improve the consistency between physical property and unit.

tomflexcompute:
Thanks @prashkh @e-g-melo !

PR #84, opened 2024-03-11
fixed MZI notebook

tylerflex:
Following the discussion here, implemented the changes and ran the notebook successfully.

However, it was working on my system before so @momchil-flex could you give this PR a run through and see if it fixes it on the env where it was broken?

https://github.com/flaport/sax/issues/31#issuecomment-1987048485


momchil-flex:
Yep works now!

tylerflex:
yay

PR #86, opened 2024-03-13
Moving adjoint case studies to case study section

tomflexcompute:
This PR moves the adjoint examples from tutorials to case studies. 

A new category "Adjoint Optimization" is added and the original "Photonics Optimization" is changed to "Evolutionary Optimization".

The thumbnail images are added to the metadata for display in the [Learning Center](https://www.flexcompute.com/tidy3d/learning-center/example-library/)


tylerflex (commented):
Thanks @tomflexcompute , the images are really great. Thanks for those!

I have a few concerns:
1. I think "Evolutionary Optimization" really only applies to genetic algorithms. I could be wrong but for particle swarm and others int his category, I think maybe "Global Optimization" would be better? 
2. I feel like splitting the adjoint examples up is a bit awkard to me.  I know we discussed putting some in the tutorials and some in the case studies, but now that I look at it, I almost feel like it should be all in it's own section. perhaps "Inverse Design" in "case studies"?  Alternatively, we can do this + have an "Adjoint Plugin" tutorial section that duplicates the first 3 notebooks (and adds a note in the landing page, telling the user to go to the "inverse Design" section for more demos?

tomflexcompute:
> Thanks @tomflexcompute , the images are really great. Thanks for those!
> 
> I have a few concerns:
> 
> 1. I think "Evolutionary Optimization" really only applies to genetic algorithms. I could be wrong but for particle swarm and others int his category, I think maybe "Global Optimization" would be better?
> 2. I feel like splitting the adjoint examples up is a bit awkard to me.  I know we discussed putting some in the tutorials and some in the case studies, but now that I look at it, I almost feel like it should be all in it's own section. perhaps "Inverse Design" in "case studies"?  Alternatively, we can do this + have an "Adjoint Plugin" tutorial section that duplicates the first 3 notebooks (and adds a note in the landing page, telling the user to go to the "inverse Design" section for more demos?

Thanks @tylerflex "Glaobal Optimization" sounds good. Regarding 2, I'm not sure if I understand what you are proposing. Could you elaborate a bit? Do you mean having some notebooks in both the case studies and tutorials?

tylerflex:
For 2, I'm imagining either

1. All of the adjoint tutorials in a single section called "Inverse Design" in "case studies"
2. All of the adjoint tutorials in a single section called "Inverse Design" in "case studies", plus
    a. A section "Adjoint Plugin" in "tutorials" containing only the `AdjointPlugin{0,1,2,4}` notebooks.
    b. Some text in the "Adjoint Plugin" landing [page](https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/docs/features/adjoint.html).telling the user that there are many more adjoint case studies in "inverse Design" section.

tylerflex:
and yea option 2 would duplicate these notebooks.

tomflexcompute:
I see. That sounds fine. However for some tutorials such as AdjointPlugin{0,1} I'm not sure what thumbnail to make. 

Maybe we just keep all the adjoint notebooks in the tutorial as they are? The thumbnail images won't be displayed but when the link is shared in social media, they will still show up. I will also rename the "FDTD Adjoint Optimization" to just "Adjoint Optimization".

tylerflex:
For 0 it could be similar to the image 
image

For 1 maybe some variation of this would be nice
image

Or we can keep them all in the tutorial section. I would also think we should consider naming it "Inverse Design" or maybe "Inverse Design / Adjoint Optimization"?

tomflexcompute:
Ok moved all adjoint notebooks back to the tutorials and renamed the section "Inverse Design / Adjoint Optimization". In the future if we decide to move things again it should be fairly straightforward.

tylerflex (approved):

daquinteroflex (approved):

PR #87, opened 2024-03-13
Updating PhC polarization splitter notebook to avoid subtracting hole geometries

momchil-flex:


lucas-flexcompute (approved):
Minor suggestion: maybe add a comment that the holes structure has to be placed **after** the silicon slab.

momchil-flex:
Done and merging, thanks!

PR #91, opened 2024-03-19
added varEps notebook

alec-flexcompute:
for some reason there're those version changes in unrelated notebooks. Is this an issue?


momchil-flex:
Ideally you shouldn't modify other notebooks even if it's irrelevant like here (to keep the history clean). It should be a matter of `git checkout`-ing these notebooks from develop, or from this branch something like:

```
git checkout -b backup/varEps
git checkout alec/varEps
git reset --hard origin/develop
git checkout backup/varEps -- img/effective_index_approximation.png
git checkout backup/varEps -- EffectiveIndexApproximation.ipynb
```

and then commit and squash.

e-g-melo (approved):
Hi @alec-flexcompute! This is an amazing example. Very useful!

My comments:
1. I suggest you to add a reference to varFDTD or 3D - 2D in the image to make it even more attractive.
2 - "This formula is based off of the assumption" -> "This formula is based on the assumption"
3. Include titles in those figures after [6], [17], and [38].
4. Include "plt.show()" in [10].
5. "Summation" -> "Summary".
6. If you have enough bandwidth, I think it would be interesting to include a third 2D simulation using only a single n_eff value obtained at the central wavelength to create the 2D model. That would let the users verify how the varFDTD results are better than this simple and popular approach.

tomflexcompute:
Thanks @alec-flexcompute will take a closer look at the notebook again shortly. In the meantime, could you add this notebook to `docs/case_studies/pic.rst`? And if you haven't done so already, could you add [metadata](https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e)?

alec-flexcompute:
Made the changes and changed the other files back to what they were. Still not sure why opening them in jupyter lab was enough to change the version metadata though.

tomflexcompute (changes requested):
Thanks @alec-flexcompute . Results are very good. Just some details that we can improve further for readability:

- The mode solving is done using the local mode solver. Could you add a comment so that users know if they really careful the accuracy of the calculated n_eff, they should consider using the remote mode solver?

- "When you mention "First we quickly build "Design 1," simply copying the values from the notebook.", could you add the link to the notebook, i.e. https://www.flexcompute.com/tidy3d/examples/notebooks/StripToSlotConverters/

- Looks like the notebook is not run from the begining so the cell numbers are not sequential, but in the current cell[7] you use `step = 6` to downsample the wavelength points. It's a good idea. Could you add comments about it so readers are aware of what it does and what the benefit is?

- Cell[18] instead of `plot`, maybe `plot_eps` can highlight the effect of 2D approximation better. Same for the other two examples.

- Cell[19] the task names for both simulations can be like "sim_3d" and "sim_2d" just to be more informative and differetiable. Same for the other two examples.

- Above cell[20] you can quickly common on the comparison. Something like "the 2D and 3D results are very similar, which validates the fidelity of the 2D effective index approach."

- I think you only need to `td.config.logging_level = 'ERROR'` once?

- Cell[66] has a warning about complex number. I think we can get rid of it by taking np.real()?

- The Y branch section title should be Example 3 now.

In general I feel like the notebook can use more comments. We should assume readers are very new to Tidy3D and this could be the first example that they read.

momchil-flex (approved):
> Made the changes and changed the other files back to what they were. Still not sure why opening them in jupyter lab was enough to change the version metadata though.

Yeah that's typical actually, notebooks are kind of annoying to manage with git. Just make sure that you're only committing the files you want to whenever you're working with them.

I'm hitting approve since I've had a quick look at the notebook and I trust Emerson and Tom's more detailed reviews.

alec-flexcompute:
Just made the changes. Let me know if I should add more comments, or make any other changes.

tomflexcompute:
Thanks @alec-flexcompute . In `plot_eps` the waveguide shows a refractive index of 1. I think this is because the material is dispersive and you didn't specify the frequency so it plots at infinite frequency. You can specify to plot at the central frequency, which should show the waveguide has a higher refractive index than the cladding. 

image

Also on second thought, the labeling in the schematic might be confusing. Maybe this one is a bit better and you can replace the current one with it. 

effective_index_approximation

Other than that I think everything looks good. Feel free to squash all commits and merge it after these final changes.

PR #93, opened 2024-03-22
Saving to current directory (instead of to output/)

marc-flex:


tomflexcompute (commented):
Thanks @marc-flex . If you just save the hdf5 to the current directory, every time you run it git might pick up the addition of the files? If so could you git ignore it? Alternatively you can save to the `data` folder, where files are ignored already. But in the notebook better check if the folder exists or not. If not, create it first as we did in cell[10] of this [notebook](https://www.flexcompute.com/tidy3d/examples/notebooks/XarrayTutorial/). This ensures when users run this notebook in there notebook web service, there will be no error.

PR #97, opened 2024-04-01
Fixed the dB scale

tomflexcompute:
In the overlap calculation, for some reason that I don't remember, I used 20*log10(overlap) to plot the overlap in dB scale. However the overlap itself should give power already so it should be 10*log10() instead?


momchil-flex (approved):
I think this makes sense, yeah.

e-g-melo (approved):

tomflexcompute:
Actually after a second thought, I'm more confused. The `outer_dot` should give the amplitude so to calculate dB it should be `20*log10()`?

momchil-flex:
Sorry. Yes, not sure why I didn't catch this before. The `dot` should be amplitude indeed.

tomflexcompute:
Ok yeah that should be the case. Let me revert the revert the change earlier.

PR #98, opened 2024-04-03
Added hollow-core fiber example

tomflexcompute:
There have been a few users interested in mode analysis for optical fiber, not necessarily doing FDTD. Since the fiber is pretty big, mode solving locally can take a long time so server mode solver might be a better option for them.

@momchil-flex the k_eff and loss are still negative. Do we plan to fix the sign in the next version? Would it be a simple fix or there are some complications"?

@e-g-melo the "thumbnail_image" path in the metadata of the structural color notebook has a slightly different format from other notebooks. It seems fine but I changed it just in case.


momchil-flex (approved):
Great! Do you know what the expected k_eff would be and if our results make sense? This could be useful for me in my mode solver PML refactor. And yest that involves fixing the sign, will be included in 2.7. Currently as long as there's no material loss the result should be correct, the sign can be ignored.

Was there a reason to use 30 pml layers? I think typically we should use 12 or say 15, like in the FDTD. But if that wasn't working well that's another thing that could give me a hint as to how best to fix the mode solver pml.

momchil-flex:
Should we hold this off until it's tested with the pml updates? Maybe we'll put it in 2.7.

tomflexcompute:
> Great! Do you know what the expected k_eff would be and if our results make sense? This could be useful for me in my mode solver PML refactor. And yest that involves fixing the sign, will be included in 2.7. Currently as long as there's no material loss the result should be correct, the sign can be ignored.
> 
> Was there a reason to use 30 pml layers? I think typically we should use 12 or say 15, like in the FDTD. But if that wasn't working well that's another thing that could give me a hint as to how best to fix the mode solver pml.

Thanks. Yeah originally I wanted to ensure PML is performing sufficiently well so used a larger number right off the bat and forgot to reduce it later. Indeed using 15 layers works pretty well. 

The keff we obtained looks quite reasonable compared to the paper. They don't exactly match but I would imagine for such a small keff it wouldn't be very easy to match the paper perfectly, as we don't know what and how they got their results either. 

For mode solving, do we actually need to use a very subwavelength grid? Or does the grid just need to resolve the mode profile and structures well?

tomflexcompute:
> Should we hold this off until it's tested with the pml updates? Maybe we'll put it in 2.7.

Yeah sure we can definitely not publish it until 2.7.

momchil-flex:
> For mode solving, do we actually need to use a very subwavelength grid? Or does the grid just need to resolve the mode profile and structures well?

Hard to say if we're talking about capturing very tiny loss... Otherwise to get the right n_eff with good precision, I wouldn't imagine a high resolution is needed.

tomflexcompute:
Just tested some different grid sizes but turns out we do need this kind of grid to get good loss value. A coarse grid makes loss artificially high.

momchil-flex:
Right, this makes sense to me for very small loss.

e-g-melo (approved):
Thanks for this great example @tomflexcompute!

I suggest you including sensing as an application example for these fibers. Maybe also include something related in SEO information.

Emerson

tomflexcompute:
> Thanks for this great example @tomflexcompute!
> 
> I suggest you including sensing as an application example for these fibers. Maybe also include something related in SEO information.
> 
> Emerson

Thanks Emerson. That's a great idea.

Some other fiber optic examples that we can add later to enrich this category:

- Step index fiber
- Graded index fiber
- Photonic crystal fiber
- Lensed fiber

momchil-flex:
Hi @tomflexcompute the mode solver PML modifications have been released in 2.7.0rc2 (solver version 24.2.2), could you try if anything changes in this notebook? The sign of k_eff should be fixed and hopefully the values should still make sense (or even make more sense or at least be more stable w.r.t. things like number of pml and the size of the mode plane, grid resolution, etc).

e-g-melo:
> PML modifications

Hi @momchil-flex! What are the modifications to mode solver PML?

momchil-flex:
- `k_eff` was coming out as negative, now fixed
- Introduced a few other modifications to make the PML more robust in various settings. Also better tested now. Discussed in detail [here](https://github.com/flexcompute/tidy3d-core/pull/599)

tomflexcompute:
The effective index and loss are nearly identical (besides the sign flip) before and after the PML change, which is good. The forth mode is not found after the change for some reason.

image

momchil-flex:
Was that an unphysical mode in the pml? If so then maybe even better?

tomflexcompute:
> Was that an unphysical mode in the pml? If so then maybe even better?

That should be a real mode. I'll play with the PML setting to see if I can get it back.

momchil-flex:
Ok, I see, let me know. Maybe we can tune even better. Does local mode solver work for this problem or is it too big? If so then you can play around with the `sigma_max` value [here](https://github.com/flexcompute/tidy3d/blob/126db25445236e271bb53999e8d8e1d48da1f906/tidy3d/plugins/mode/derivatives.py#L199).

tomflexcompute:
hmm tried a few things but don't seem to find the 4th mode. Maybe it doesn't have to do with PML but other mode solver changes from 2.6 to 2.7? That mode usually is not what people are looking for but should be a physical mode I think. Not sure if we should be worried about it.

tomflexcompute:
Yeah so I used 2.7rc1 to run this notebook and the fourth mode is missing as well. Some mode solver changes between 2.6 and 2.7 before the sign and PML change was causing the difference here.

momchil-flex:
I think this is related to @weiliangjin2021 's subspace change then. I tried with the commit right before that and I got the four modes. However, I also get the fourth mode if I compute for *five* modes with the new solver. Then the fifth one is missing.

I don't know if @weiliangjin2021 wants to debug more... It's a bit hard because this is too big to run locally. It also seems kind of OK to me as is but maybe pointing to potential problems down the line?

tomflexcompute:
Ok now solving for 5 modes work. The loss numbers are also correct. Should be ready to merge into the 2.7 branch. If there are no additional comments I'll merge it later today.

PR #99, opened 2024-04-04
Added EMESolver notebook

caseyflex:


momchil-flex:
Really nice! Some things I noted when playing around:

- We should validate that `symmetry=0` along the eme propagation axis.
- In the taper example, in the FDTD case, we actually use a gaussian beam source. Here as far as I can tell you are actually looking at just the propagation of the fundamental mode, i.e. your first mode solve is already *inside* the waveguide. Interestingly, despite of that the coupling efficiency at large taper length seems to be higher than that of FDTD, even though the latter may include some reflection from the gaussian beam into the interface. Is this due to the EME step size in the propagation direction, or something else? Have you tested?
- I am not sure about the extra confusion that the `port_offsets` add. Specifically, this forces a splitting of the grid definition: e.g. your first subgrid boundary is defined as `1.001*pad_x` because later on you add a `pad_x` offset. Do these offsets have any practical significance apart from nicer visualization in some cases?
- The length sweeping is great. But it does seem like we also need to have the option to only scale a single subcell. For example in the MMI we may want to scale the MMI region without scaling the in/out tapers. What's your thinking there?
- `If too many modes are used, it is common to see a warning about invalid modes.` This warning would only appear in the simulation log right? If so then say so since it will not be shown by default.
- Having a quick explanation about the change of basis is good, but we really should showcase it e.g. in the MMI. Are you planning to prepare this example for the official release?

caseyflex:
> Really nice! Some things I noted when playing around:
> 
> * We should validate that `symmetry=0` along the eme propagation axis.
> * In the taper example, in the FDTD case, we actually use a gaussian beam source. Here as far as I can tell you are actually looking at just the propagation of the fundamental mode, i.e. your first mode solve is already _inside_ the waveguide. Interestingly, despite of that the coupling efficiency at large taper length seems to be higher than that of FDTD, even though the latter may include some reflection from the gaussian beam into the interface. Is this due to the EME step size in the propagation direction, or something else? Have you tested?
> * I am not sure about the extra confusion that the `port_offsets` add. Specifically, this forces a splitting of the grid definition: e.g. your first subgrid boundary is defined as `1.001*pad_x` because later on you add a `pad_x` offset. Do these offsets have any practical significance apart from nicer visualization in some cases?
> * The length sweeping is great. But it does seem like we also need to have the option to only scale a single subcell. For example in the MMI we may want to scale the MMI region without scaling the in/out tapers. What's your thinking there?
> * `If too many modes are used, it is common to see a warning about invalid modes.` This warning would only appear in the simulation log right? If so then say so since it will not be shown by default.
> * Having a quick explanation about the change of basis is good, but we really should showcase it e.g. in the MMI. Are you planning to prepare this example for the official release?

- I actually just added this simulation-level validator for symmetry in propagation direction.
- Yeah, I am just looking at propagation of the fundamental mode. Interestingly, when I use this composite grid with the small cell on either end, I get lower efficiency, but when I use a uniform grid I get higher efficiency as expected. I think this has to do with subpixel at the ends of the waveguide -- when the cell is extremely small and not aligned to Yee grid, the subpixel can lead to weird modes, but when the cell is a reasonable size or aligned to Yee grid it should be ok. For now, I'll get rid of the small cells on either end -- it would be good to add option to make Yee grid generate conformal to EME cells. I'll add a TODO comment about this
- Prash and Lucas both wanted something like this so that the ports didn't have to be exactly at the simulation boundaries. I can simplify the setup in the demo notebook, it isn't usually this convoluted -- I was just trying something to reduce storage of port_modes
- Yeah, I was thinking of supporting `scale_factor` being an array indexed by `cell_index` so that specific cells can be scaled.
- done
- The MMI could be a separate case study, rather than appearing in the basic EME solver demo notebook

daquinteroflex:
Looks great, docs appear well integrated too!

Tiny suggestions:
- [x] Maybe it's worth adding the tips and tricks section somewhere in the `Notes` heading of the `EMESimulation` API class

Maybe not for this notebook:

If it's meant to be an intro for new users to the EME method, it'd be handy to explain the forward and backward propagating mode superposition aspect of the solver implementation. Just like the maths behind it maybe as you did in your presentation beforehand. Like literally including parts of that presentation into the guide would probably be really interesting for new users evaluating the solver accuracy, which one to use, etc.

daquinteroflex (approved):

caseyflex:
> Looks great, docs appear well integrated too!
> 
> Tiny suggestions:
> 
> * [x]  Maybe it's worth adding the tips and tricks section somewhere in the `Notes` heading of the `EMESimulation` API class
> 
> Maybe not for this notebook:
> 
> If it's meant to be an intro for new users to the EME method, it'd be handy to explain the forward and backward propagating mode superposition aspect of the solver implementation. Just like the maths behind it maybe as you did in your presentation beforehand. Like literally including parts of that presentation into the guide would probably be really interesting for new users evaluating the solver accuracy, which one to use, etc.

thanks! added tips and tricks to `EMESimulation` API class, and added a brief discussion about EME maths to the notebook. We could have a longer discussion elsewhere later

prashkh (approved):
Notebook looks great! I really like the tapered sweep example. 

Here are a few suggestions:
1. Having a few schematic images in the beginning showing what EME is good for and what it is not good for might be instructive. 
2. For tapered waveguide cross-section plot, it was hard to see the pink color of the tapered structure. Maybe you can make this obvious with a short taper?
3. If you could have a plot showing the different modes eme solver calculated at a particular eme cell, that would be instructive as well.
4. It would be great if you could plot a 2D image showing the scattering matrix between the input and the output port modes, i.e. visually seeing the diagonal and off diagonal components would be nice. People use this to inspect which eme cell is causing most amount of scattering.

Congratulations on the upcoming release of the EME solver!

momchil-flex (approved):

PR #100, opened 2024-04-04
Fixed numpy deprecation warning on converting array with single eleme…

tomflexcompute:
Found two instances where there is a warning on converting an array with a single element to a scaler. Fixed them in this PR to avoid future error when switching to numpy 2.0.


daquinteroflex:
Thanks for this Tom and raising this!

Hmm this is quite interesting. 

We've fixed numpy in our requirements so they shouldn't fail exactly, but I suspect maybe when we do actually do the upgrade we'll have to go through a fair bit of these notebooks to check, or maybe little by little the warnings will help us maintain compatibility. I guess this is why we need the notebook testing too.

daquinteroflex (approved):

tomflexcompute:
Thanks @daquinteroflex yeah they shouldn't error any time soon but I figure we could save some time and hassle if we fix it sooner rather than later. Also getting rid of the warning makes the documentation look cleaner.

PR #102, opened 2024-04-09
fix files and paths causing issues in MC notebook runs

tylerflex:
The file in `misc` is different from the fname in the QE extractor notebook, causing problem (3) in https://flow360.atlassian.net/browse/SCEM-5706

Also, sometimes the `to_gds_file()` gets called but `misc` directory is not present, added code to mkdir it if it doesn't exist in some notebooks.

```py
# make the misc/ directory to store the GDS file if it doesnt exist already
import os
if not os.path.exists('./misc/'):
    os.mkdir('./misc/')
```

### DIFF
--- a/AdjointPlugin12LightExtractor.ipynb
+++ b/AdjointPlugin12LightExtractor.ipynb
@@ -1136,7 +1136,7 @@
    "outputs": [],
    "source": [
     "# where to store history\n",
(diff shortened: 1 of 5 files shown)


e-g-melo (approved):
Thanks @tylerflex!

I will include the `os.mkdir('./misc/')` in the next notebooks.

PR #104, opened 2024-04-18
adding coupled line bandpass filter as a microwave case study

dmarek-flex:
First draft of my first completed notebook. This example was originally created by @tomflexcompute which I have heavily modified and extended. I am using the new lumped ports that are part of microwave additions to the smatrix plugin.

Some concerns:

- I think I am using additional Python packages that might not be part of the default import list (scikit-rf and maybe SciPy). Does that need any additional changes?
- I was going to use this notebook to introduce the TerminalComponentModeler, but it is probably better to create a simpler dedicated tutorial for that which I will be doing in the coming days.


tomflexcompute (changes requested):
It's fantastic to see the first microwave/RF example coming out! It has been quite a while since I made the initial version of the notebook. The current notebook already looks great. Thanks @dmarek-flex for the great efforts. Some minor and formality comments:

- In our convention, the notebook title only capitalizes the first character of the first word. 

- Besides the title, in other section headers we do capitalize the first character of every word (not sure why we follow this convention but it started off this way early on so we just keep it through). However section headers need to be lower levels, e.g. "## Filter Design and Microstrip Parameters" otherwise the docs will not be built correctly, at least it used to be the case and Dario can confirm. 

- In microwave.rst, the length of "~" needs to match the length of "Microwave" otherwise the docs will also not be built correctly. At least that used to be the case and Dario can confirm if that's still necessary now. 

- Instead of just microwave, maybe we can make the title "Microwave and RF Devices" for better SEO. 

- When referring tidy3d classes in the markdown such as TerminalComponentModeler, it would be good to add the link to the API reference.

- For RF applications, I suppose the warning (cell[7]) about spacing smaller than half a wavelength will always show up. Not sure if we should set the logging level to error to get rid of the warnings. If you decide to do so, please add a comment about the reason behind it. 

- After plotting, add "plt.show()" to avoid printing out additional information such as Text(0.5, 1.0, 'Side view of port 2').

- The optimization is very nice. I suppose a somewhat large number of simulations were done to obtain the optimized result. Could you comment on the total number of simulations, total simulation time, as well as a brief introduction of L-BFGS-B?

- Some typos: 
"charateristic" -> "characteristic"
"mesh overrides structures" -> "mesh override structures"

- We also typically add a thumbnail schematic at the beginning of the notebook to quickly show the reader what this notebook is about. The same thumbnail image will be used in the [learning center](https://www.flexcompute.com/tidy3d/learning-center/example-library/). You can use this image I just created. 

coupled_line_filter

- Lastly we need to add metadata to the notebook for SEO purposes. You can follow the steps [here](https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e) to do so.

It could seem like a good amount of information so definitely feel free to let me know if anything is unclear.

dmarek-flex:
Thanks for the advice, see my comments below.

> * For RF applications, I suppose the warning (cell[7]) about spacing smaller than half a wavelength will always show up. Not sure if we should set the logging level to error to get rid of the warnings. If you decide to do so, please add a comment about the reason behind it.

Yes, I always get a lot of warnings like that! I will change the logging level unless someone as has a better idea.

> * The optimization is very nice. I suppose a somewhat large number of simulations were done to obtain the optimized result. Could you comment on the total number of simulations, total simulation time, as well as a brief introduction of L-BFGS-B?

The optimization step actually does not involve any Tidy3D simulations. It uses functions I added to the microwave plugin that compute transmission line parameters using approximate models found in the literature. I will clarify this in the notebook.

daquinteroflex:
Looks great, and really interesting read too! It's nice to show that these types of structures can be layout-automated like you have. Just general questions:

- Not for this example exactly, just thinking in general for the API. I wonder how to make it easier to specify the mesh overrides in this case. Is what Weiliang is working on currently is an automated way to do the mesh overrides so that they're easier to define for the user or done automatically? I guess I'm also thinking how to make it easier for users to check them as it currently involves some specific hlims/vlims etc, but maybe because they're inherently application specific it is derived.
- The scikit-rf validation is great for people wanting to prove the validity of the simulation setup. I wonder if a few more warnings about the importance of mesh overrides, etc. in these type of simulations is worth it, or adding some of this information within the API code documentation itself. We can have a dedicated rst page for common RF simulation best-practices possibly? Think it might be worthwhile.
- Really like how you've proven the optimisation.
- Even though the insertion loss method is more detailed, I believe a bit more of an intro overview in terms of the tradeoffs required in terms of filter design and the general goal of the example might be pretty good for people who haven't used it before. Maybe a generic diagram like this might be nice for them:

image


- Docs look good! Also agree with Tom on his suggestions.

Congratulations on the great effort! Just brainstorming on ideas here.

weiliangjin2021:
Nice notebook!

- worth explaining why `remove_dc_component=True` is needed?
- in port refinement `td_port_cells = 2`, add explanations on the typical rule of thumb?
- `We turn off subpixel in this example, since the structures are rectilinear and well-resolved by the grid`: there is no harm turning it on? And currently, the default subpixel=True is in fact the same as subpixel=False for PEC when conformal_spec is not set. I think we can use subpixel=True and the default conformal_spec.
- Regarding mesh override structures, @daquinteroflex you are right, we are working on more automatic generation. Now in this notebook, `dl=[total_L / 200, np.min(widths) / 20, h_trace]`, are the grid size along x/y-axis necessary?

dmarek-flex:
OK, I believe I have addressed each of your comments. Some specific responses:

- @daquinteroflex @weiliangjin2021  I'll leave the discussion of mesh refinement for when the automated feature is released. I tried without the mesh overrides in the x and y direction and I still got fairly good results, but definitely more accurate results when I used the mesh overrides.
- @daquinteroflex Instead of a diagram, I just organized the explanation into bullet points with some additional info.
- @weiliangjin2021 I reverted to defaults to avoid needing to explain them. Actually by using the default for port_refinement, the results got a little closer to the scikit-rf results.

daquinteroflex (approved):

tomflexcompute:
Looks great! I think we will wait until 2.7 is released to merge this PR?

dmarek-flex:
> Looks great! I think we will wait until 2.7 is released to merge this PR?

Yes, the TerminalComponentModeler itself was just added in 2.7 I believe

weiliangjin2021:
> > Looks great! I think we will wait until 2.7 is released to merge this PR?
> 
> Yes, the TerminalComponentModeler itself was just added in 2.7 I believe

In that case, why not merge into pre/2.7 branch?

weiliangjin2021 (approved):

dmarek-flex:
> > > Looks great! I think we will wait until 2.7 is released to merge this PR?
> > 
> > 
> > Yes, the TerminalComponentModeler itself was just added in 2.7 I believe
> 
> In that case, why not merge into pre/2.7 branch?

Yea sorry that makes sense. I'm probably not the best person to ask about releases!

dmarek-flex:
@tomflexcompute Kind of forgot about this, but may I merge into pre/2.7?

tomflexcompute:
> @tomflexcompute Kind of forgot about this, but may I merge into pre/2.7?

Yeah definitely feel free to merge it.

PR #105, opened 2024-04-23
Added a mid-ir metalens notebook and associated files

tomflexcompute:
Quite a few users have been asking about how to design a metalens/meta-diffractor from scratch, i.e. from unit cell simulation to the full metasurface construction. I have had this example for a while mainly as a test for large metalens simulation but never got to release it. After discussing with @e-g-melo we decided to release it as a public example. Hope this example serves as a tutorial for that and helps answer some of the questions. 

Also fixed the title capitalization for the adjoint quick start notebook.


tomflexcompute:
I demonstrated the gds export at the end of the notebook. It works very well in this case. However it reminds me that a user was trying to do gds export for a large metalens (millions of cells) and the export took over 7 hours. They were wondering if our gds generation algorithm could potentially be further optimized for speed @lucas-flexcompute ?

lucas-flexcompute:
Our gds export depends on intersection calculations done with shapely, so, in general, hard to optimize.
That said, a particular geometry might be easier to handler than the general case, but that depends on building a specific exporter for that particular case.

tomflexcompute:
> Our gds export depends on intersection calculations done with shapely, so, in general, hard to optimize. That said, a particular geometry might be easier to handler than the general case, but that depends on building a specific exporter for that particular case.

The only problematic case is probably large metalens/metasurface where you have tons of nonintersecting structures. Might be worth it to have an optimized exporter for this case specifically.

lucas-flexcompute:
Yes, those are still a general case. What I meant was more something like creating the gds manually from the optimized parameters. For example, if the metalens design is based on rotated, identical elements, create the gds with a base element and thousands of references using the optimized rotations.

momchil-flex:
In this example, all geometries are Cylinders. There must be a very efficient representation in GDS in which maybe the cylinder radius is quantized up to some value, and then you just need as many cylinders as discrete values of the quantization, plus many references to those. Right? But the only way to use this fact that I can think of currently is to kinda invert the problem, and write a GDS script that makes the metalens first, and then import that in Tidy3D.

e-g-melo (approved):
Hi @tomflexcompute! Thanks for this example. It is very didactical.

I found only a typo: "and the periodic of the unit cell is fixed" -> "and the periodicity of the unit cell is fixed"?

No more comments!

lucas-flexcompute:
> In this example, all geometries are Cylinders. There must be a very efficient representation in GDS in which maybe the cylinder radius is quantized up to some value, and then you just need as many cylinders as discrete values of the quantization, plus many references to those. Right? But the only way to use this fact that I can think of currently is to kinda invert the problem, and write a GDS script that makes the metalens first, and then import that in Tidy3D.

Yes, that's right. In this case, instead of a Cartesian array of cylinders, you could also use a cylindrical array to make use of symmetry. The GDS creation and import can be done only after the optimization, just to make sure the discretization, quantization, and the actual script are working.

Since we're discussing this example in particular, I'd also try to use an non-periodic array with more homogeneous density than a cylindrical array, which helps eliminating periodic artifacts. I've used the [Fermat spiral with a golden angle](https://en.wikipedia.org/wiki/Fermat%27s_spiral#The_golden_ratio_and_the_golden_angle) before and got excellent results.

momchil-flex:
More practically do you think if we don't use our in-built gds export, but e.g. iterate over the cylinders using a small script and gdstk directly, we could have a much faster export for a large structure? Even if we don't use references?

lucas-flexcompute:
Yes, that's also probably faster. The final file will be quite large, though.

momchil-flex (approved):

PR #106, opened 2024-04-25
uncomment accessing batch data in Design.ipynb

tylerflex:
There was some note about batch data only accessible > 2.6. Safe to remove it now.


tomflexcompute (approved):
For some reason I couldn't see the difference, but the notebook looks good!

PR #107, opened 2024-05-03
Minor updates to adjoint notebook removing `to_simulation()` calls where possible

yaugenst-flex:
Updates a few notebooks to remove `JaxSimulation.to_simulation()` calls where safe to do so, particularly before `to_gds()` calls as this is fixed as of flexcompute/tidy3d#1660.
The changes are very conservative, I did not remove `to_simulation()` calls if the resulting simulation object is being reused for other things, such as running a simulation.


tylerflex (approved):
Looks good, thanks for taking care of this! (there might be a few other instances, eg in the mode converter but it's ok, we can just clean them up when we notice them)

PR #109, opened 2024-05-13
Update for prefab 1.0.0

Dusandinho:
@tylerflex We made some significant changes on our end (prefab v1.0.0) that require an update to this notebook:
- Device designs are loaded and handled differently
- Models are handled differently (from a library)
- Models have been updated to new data

This update makes the necessary changes. Because the models have been updated, the results are very slightly different.


tylerflex (approved):
Prefab 1.0.0? First official release? Exciting! :)

Dusandinho:
> Prefab 1.0.0? First official release? Exciting! :)

Still lots of room for growth and improvement but I think the right fundamentals are now in place.

PR #112, opened 2024-05-17
add notebook with basic microwave simulation using the impedance calculator

dmarek-flex:
This is a very introductory style notebook to explain some of the functionality in the microwave plugin.


dmarek-flex:
Feel free to take a look @weiliangjin2021 and @QimingFlex. I just need to make some final touches, but the content of the notebook is ready for review.

dmarek-flex:
Ok just missing a rendering. Also not sure how we want to organize this in the docs. I called it a feature tutorial under the "microwave" folder, since it uses a lot of functionality from the microwave plugin.

tomflexcompute (approved):
Very good tutorial. Thanks @dmarek-flex ! For feature tutorials we typically don't need to have a 3D rendering. A few minor comments:

1. The time-domain field in the last plot is very nice. It might be interesting to also plot a frequency-domain reflection spectrum to see the reflection is minimized at the central frequency when the load is matched? 

2. "**Setup** structures and simulation..." -> "**Set up** structures...". Same for "we need to **setup** the Tidy3D simulation..." and "Now, the simulation is **setup**...".

3. In microwave.rst, maybe make the title something like "Microwave and RF" for better SEO?

weiliangjin2021:
Quite enjoyed going through the tutorial, thanks! A few comments below:

- Can the substrate size in xy be set to `td.inf`?
- It's fine to use UniformCurrent in this tutorial, but I think it would be helpful to add a cell showing that it's also possible to automatically generate a suitable current source from a lumped port.
- In the 1st simulation, the transmission line terminates in PML? If so, it would be helpful to describe that in the session "Setting up a Microstrip Simulation". When you define `run_time`, there is a description "Fields should decay rapidly in this notebook, since we are using PMLs and matched loads". So I was expecting a lumped resistor.
- In session "Computing the Characteristic Impedance", add some explanation below the figure that compares tidy3d and quasi-static model? E.g. how much difference is expected between the 2 curves?
- In session "Defining a custom path integral", the figure looks identical to the previous figure? How will the value of `Nx, Ny` affects the result? I guess they cannot be too small as the numeric integration might not converge. Then what would be a good value here? And what will happen if their value is too huge?

dmarek-flex:
I finished making the changes you both requested.

> 1. The time-domain field in the last plot is very nice. It might be interesting to also plot a frequency-domain reflection spectrum to see the reflection is minimized at the central frequency when the load is matched?

I experimented with this, but the plots were not very illuminating because the incident field is also recorded by the frequency domain monitor. I'll leave that type of analysis for the tutorial on the smatrix plugin.

> In session "Computing the Characteristic Impedance", add some explanation below the figure that compares tidy3d and quasi-static model? E.g. how much difference is expected between the 2 curves?

I added a brief discussion about the accuracy. I also did a lot of tests to see what I could do to increase the accuracy of Tidy3D. The most effective changes was increasing the grid resolution. The final accuracy is on the order of 1% which I think is reasonable.

daquinteroflex:
Hello! I also really enjoyed going through this notebook. I liked the way you connect the theory with the simulations, the figures, and how progressive it feels to get familiar with the features. I like how you go from the axis-to-custom voltage/current integral calculations and relate it back to the theory.

In terms of the docs, looks correct to me how you've done it!

daquinteroflex (approved):

weiliangjin2021 (approved):

QimingFlex:
@dmarek-flex Very well-prepared notebook, and I enjoyed reading it as well! A few comments:

    1. It seems the PML is quite far from the circuit port along the x-y plane. In such a case, the port will be considered open. I think the PML should be directly connected to the edge of the microstrip to absorb waves.

    2. The diagram is beautiful! However, there are some differences between the parameters shown in the schematic diagram and the actual parameters. For example, the thickness of the substrate is 2*height in the simulation but just height in the diagram.

    3. Would it be better to plot voltage and current in real units like V and A? This way, users can easily see the impedance by comparing the peaks.

QimingFlex (commented):
@dmarek-flex  Very well-prepared notebook, and I enjoyed reading it as well! A few comments:

- It seems PML is quite far from the circuit port in the x-y plane. In such a case, the port will be considered open. I think PML should be directly connected to the edge of the microstrip to absorb waves.

- The diagram is beautiful! But it seems there are some differences between the parameters shown in the schematic diagram and the actual parameters. For example, the thickness of the substrate is 2*height in the simulation but just height in the diagram.

- Would it be better to plot voltage and current in real units like V and A? This way, users can easily see the impedance by comparing the peaks.

dmarek-flex:
> * It seems PML is quite far from the circuit port in the x-y plane. In such a case, the port will be considered open. I think PML should be directly connected to the edge of the microstrip to absorb waves.

The problem there is then the excitation would be inside the PML. I could have also added a matched load at the incident side, but I wanted to keep it as simple as possible.

> * The diagram is beautiful! But it seems there are some differences between the parameters shown in the schematic diagram and the actual parameters. For example, the thickness of the substrate is 2*height in the simulation but just height in the diagram.

The thickness is indeed 2x the height, but the box is centered at z = 0, so the effective height in the simulation is correct. I did that to avoid warnings in Tidy3D when objects are too close to simulation boundaries.

> * Would it be better to plot voltage and current in real units like V and A? This way, users can easily see the impedance by comparing the peaks.

The reason I normalized the voltages and currents is because the absolute values I get are not pleasant. If I remember correctly the voltage is in the tens of thousands of volts, so I think that might be confusing! We would need to change how the UniformCurrentSource is normalized in order to get a more reasonable voltage/current. Incidentally, the default use of microns instead of the `mm` is a similar problem and we should fix this at some point.

QimingFlex:
> > * The problem there is then the excitation would be inside the PML. I could have also added a matched load at the incident side, but I wanted to keep it as simple as possible.
> Can we move the source a bit inward towards the strip so it won't touch the PML? Reflections will be generated at the port, which might affect the accuracy of the characteristic impedance computation.

Everything else looks good!

QimingFlex:
> * The problem there is then the excitation would be inside the PML. I could have also added a matched load at the incident side, but I wanted to keep it as simple as possible.
>   Can we move the source a bit inward towards the strip so it won't touch the PML? Reflections will be generated at the port, which might affect the accuracy of the characteristic impedance computation.

Everything else looks good!

QimingFlex:
> The problem there is then the excitation would be inside the PML. I could have also added a matched load at the incident side, but I wanted to keep it as simple as possible.

Can we move the source a bit inward towards the strip so it won't touch the PML? Reflections will be generated at the port, which might affect the accuracy of the characteristic impedance computation.

 Everything else looks good!

dmarek-flex:
> Can we move the source a bit inward towards the strip so it won't touch the PML? Reflections will be generated at the port, which might affect the accuracy of the characteristic impedance computation.

On your first point, it is not really possible because the source is not directional. If I placed the source somewhere else besides the end of the strip, it will excite fields propagating in both directions. 

For your second concern, this should not be an issue since in the first simulation where I compute impedance, the PML will not produce any reflections (or at least they are very small). You can see for yourself in the first plots of voltage and current in the time domain, where there are no reflections.

PR #113, opened 2024-05-20
Added free-form coupler notebook and associated files

tomflexcompute:
This is adapted from a customer demo. The curved coupler surface is made in CAD and imported into Tidy3D since it's hard to construct such a geometry natively. The notebook itself is fairly straightforward otherwise.


e-g-melo (approved):
Thanks @tomflexcompute! It is a very nice structure.

- I am not sure, but it seems that the text was not formatted using black.
- The text refers to "Coupling Efficiency", and the chart shows "Insertion Losss". We should unify text and chart.
- Something unrelated to the notebook: When two structures face each other, this thin line is confusing. I think we should change the plot function to show a continuous structure.

image

tomflexcompute:
Thanks @e-g-melo ! All great points. I felt like I forgot to do something and turns out it's reformatting the code! Formatting has been applied nad the terminology has been unified now. 

Regarding the tiny gap in the plot, I totally agree it's quite annoying and appears in many notebooks. Unfortunately it's deeply rooted in matplotlib and the resolution of the plot so would be difficult for us to fix. Hopefully at some point we can come up with a fix with some customization.

alec-flexcompute:
Looks good @tomflexcompute! The only thing I'm thinking about is using Tidy3D's design plugin, but that's a nontrivial reworking of this example and this type of parameter sweep is totally fine as is.

alec-flexcompute (approved):

tomflexcompute:
Thanks @alec-flexcompute . I was thinking about using the design plugin but for a simple sweep like this, I feel like using the design plugin is more work than using a simple batch so I just went with it. For more complex optimization the design plugin is recommended. I'll add a comment about it in the notebook and link to some tutorials for SEO.

alec-flexcompute:
Sounds good, other than that this looks great!

PR #116, opened 2024-06-03
Update autograd notebooks: Remove more JAX references, fix imports

yaugenst-flex:
This removes some more references of jax from the autograd notebooks and fixes `tidy3d.plugins.autograd.invdes` imports.


tylerflex (commented):
Thanks @yaugenst-flex. This was using `pre/2.7` right?

It's a little weird to me that the objective function decreases at the end in this new run (notebook 3). Hopefully not anything concerning?

image

Guess we should probably run the grating coupler through with no keyboard interrupt. but maybe we can do that while the final testing is going on for 2.7

yaugenst-flex:
Yeah I noticed that too, need to look into it a bit more. This PR is mostly cosmetic though.

tylerflex:
could be some mismatch between the penalty and physical gradients.

PR #123, opened 2024-06-14
inverse design with bilayer grating coupler

tylerflex:


e-g-melo (commented):
Thanks @tylerflex! The simultaneous optimization of different design regions is very interesting and shows the power of our inverse design. It is a very cool example.

My comments:
1. We could use this opportunity to include an introductory text summarizing the differences between the new inverse design approach using `autograd` and the older one using `jax`. An introduction to the design problem is also important. Even though you have included a link to that paper, most users might simply move to the next example instead of taking the time to read the paper.
2. Some comments on "Inspecting Mode Solver" would help users to understand which mode you are looking for.
3. Even though the simulation workflow is very similar to that "dispersive metallic antenna", users can open this notebook without reading any other one, so including more comments would help them understand the inverse design setup and bring their attention to all these nice and powerful new features.

tomflexcompute (commented):
Thanks @tylerflex for working on this on short notice. Great results! Agree with Emerson that adding more general markdown comments can help the readability especially for users who are new to it. I can add some comments and push to this branch if you prefer. Additional minor feedback:

1. Might consider following the naming convention and naming this one something like `Autograd15TwoLayerCoupler`?
2. Why the colormap for different mode field components are different?

tylerflex:
Thanks for both of your comments. I addressed them all in my PR

e-g-melo (commented):
Thanks @tylerflex!

I have only two additional comments:
- "Net we'll define" -> "Now, we'll define" ?
- It is showing this error `NameError: name 'np' is not defined` after cell [10]

tylerflex:
Thanks @e-g-melo . I fixed the typo,
The 2nd one happened because I re-ran the notebook accidentally. We'll be sure to re-run it before the release and it will work.

PR #124, opened 2024-06-14
inverse design with dispersive metallic antenna

tylerflex:
Requires https://github.com/flexcompute/tidy3d/pull/1761 for `CustomPoleResidue` support in autograd.

Also the results aren't great yet.

image


e-g-melo (commented):
Thanks @tylerflex! The new inverse design plugin seems very exciting!

Some comments about the notebook:
1. Maybe @tomflexcompute could create a nice image for this example?
2. I suggest to make this sentence more clear. "This uses the new autograd support in tidy3d >= 2.7.0 but the same differentiation support with respect to td.CustomPoleResidue is not supported in tidy3d's adjoint plugin.".
3. Merge cells [2] and [3].
4. It seems there are some comments in the cell [8] you need to remove. Like "# doesnt work" and "# medium_gold =  td.material_library['SiN']['Horiba']"
5. Cell numbers are missing after cell [11].
6. It would be nice to introduce and give more details on the `invdes` plugin and the functions `make_filter_and_project`,  and `get_kernel_size_px` after cell [11]. Or maybe include more details about the new inverse design architecture using `autograd` in the introduction. So, the existing Adjoint users could easily understand the differences between the new approach and the older one using `jax`.
7. Some typos:
- "the desity value" -> "the density value"
8. It seems `isbox` is imported but not used. If that is not the case, we could have some comments about it.
9. Should we show the initial antenna geometry in the xy plot?
image
10. I didn't find the definition for `sim_antenna_random`.
11. There are some empty cells along the notebook.

tylerflex:
@e-g-melo I incoporated your suggestions, thanks!

@tomflexcompute would you mind taking a look at this updated notebook? the results are much better now.

tomflexcompute:
> @e-g-melo I incoporated your suggestions, thanks!
> 
> @tomflexcompute would you mind taking a look at this updated notebook? the results are much better now.

Thanks @tylerflex . I remember leaving comments on both PRs but not sure why it's gone on this one. Anyway happy to read over the notebook again.

e-g-melo (approved):
Thanks @tylerflex! It is a very interesting example!

PR #126, opened 2024-06-18
Added MMI demo of EME solver

tomflexcompute:
This is the demo of using EME to design an MMI MUX. More specifically to demonstrate the length sweep and `smatrix_in_basis` function.


e-g-melo (approved):
Hi @tomflexcompute! This example is fantastic! Thanks for sharing. 

Some comments:
1. For this sentence: "Large MMI components can be expensive to design with FDTD simulations alone, but the eigenmode expansion (EME) method is much more efficient because it doesn't require meshing the multimode region.", what do you think about including something more positive for both EME and FDTD? Maybe something like "The eigenmode expansion (EME) is very efficient for designing large MMI components because it doesn't require meshing the multimode region. Even though our FDTD solver is very fast, using EME is advantageous in some situations. For example, when you are interested in just a few wavelengths, simulating very long waveguides using EME can be much faster than FDTD, although sometimes the results can be less accurate."
2. "# initial length of the mm" -> "# initial length of the mmi" 
3. "First, calculate the plot the transmission" -> "First, calculate and plot the transmission"?

To my understanding: Why have you used the `ModeSolverMonitor` instead of the `EMEModeSolverMonitor`?

tomflexcompute:
Thanks @e-g-melo ! Feedback has been incorporated. 

And that's a great question. I actually don't remember why but vaguely remember `EMEModeSolverMonitor` didn't do what I was looking for. @caseyflex do you think that's the case?

caseyflex:
> Thanks @e-g-melo ! Feedback has been incorporated.
> 
> And that's a great question. I actually don't remember why but vaguely remember `EMEModeSolverMonitor` didn't do what I was looking for. @caseyflex do you think that's the case?

Yeah, with the current terminology, the `EMEModeSolverMonitor` directly records the modes used for the EME expansion -- it doesn't perform an independent mode solve but merely records EME (super)modes within its geometry. While the `ModeSolverMonitor` behaves the same as in FDTD, solving for the modes inside its geometry, and the result can then be used to change basis for transmission to a particular output waveguide. I'm open to suggestions for future changes to this terminology, although we are stuck with this for the current 2.7 release.

tomflexcompute:
That's right. Now I recall. I think I had the same confusion as Emerson when I first tested it. 

I think the current workflow works very well but just the naming can be slightly confusing. Adding a note on the `EMEModeSolverMonitor` documentation [page](https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.EMEModeSolverMonitor.html) to explain the difference between `EMEModeSolverMonitor` and `ModeSolverMonitor` could be very helpful.

caseyflex (approved):
Thanks Tom, I think this looks great and is an exciting demo of EME, especially the ability to simulate very long structures and perform efficient length sweeps, and the usage of the smatrix_in_basis function together with `ModeSolverMonitors`.

PR #128, opened 2024-06-25
added LNOI PSR notebook

alec-flexcompute:


e-g-melo (approved):
Hi @alec-flexcompute! Thanks for this example. Awesome project!

I have only a few suggestions:
1. Considering the image, at first glance, I thought that both input modes TE0 and TM0 were split at the y-branch. Maybe there is a way to differentiate the arrows to clearly indicate the mode paths. 
2. Include a H2 header after cell [1]. Maybe "Simulation Setup"?
3. It seems the code formatting is different from the other notebooks. I wonder if you have used `jblack`.
4. I suggest to include comments to explain the charts ploted in [27].

prashkh:
@alec-flexcompute The example notebook looks great! Perfect example to demonstrate EME simulation. 

Here are my comments:
a. Annotate w0, w1, L1, L2, L3, etc in your 3D image for clarity. This way users can see all the different knobs they have for optimization. 
b. Can you also label port 1 and port 2 somewhere? 
c. Have you investigated why we are we not seeing close to 100% mode conversion in the plot "mode conversion efficiency vs L2". I see that FDTD and EME results match for the resolution you have. Since breaking the vertical symmetry with the angled waveguide is critical to achieving this TE1 and TM0 crossing, I wonder if the current transverse resolution is not fine enough?
d. The dips in mode conversion efficiency for TE0 looks like simulation artifact. Any ideas what could be causing this? Might be better to resolve them. You are using an anisotropic medium so have you checked PML vs absorber for the boundary conditions?
e. In the final plot for TM0 input, why is TM0 in port 1 so large (~-3dB) close to 1500 nm. This is a really large cross-talk that was not simulated (Figure 4d) or measured experimentally (Figure 6b) of the paper. I would try to investigate this before publishing the notebook.

alec-flexcompute:
Thanks @e-g-melo and @prashkh!

So for c, this was a source of investigation and we didn't find a reason why conversion didn't go to 100% I believe the transverse resolution didn't have much of an effect, but that was a while ago. I'll rerun to double-check.

For d, this has been a source of noise that is known and I believe @caseyflex ran into as well. What's weird is that the number of these unphysical dips tend to increase as the number of cells increase, and the actual sweep values don't seem to have an effect. If you have any ideas as to what's causing this I'd be very interested.

For e, we're not sure why there's cross-talk here. However, this PSR is using different length values than in the paper/what was fabricated

caseyflex (changes requested):
Thanks @alec-flexcompute for the example notebook. I think it's a really exciting demonstration from both EME and FDTD sides.

I do think we should understand the agreement with the paper a bit better before publishing. It looks like our EME and FDTD solvers are agreeing, but the paper is somewhat different, for example in Fig. 2 (b). It would be good to double check everything -- grid resolution, run_time, material properties, geometries, subpixel, PML, etc. It looks like we should get complete conversion from TM0 to TE1 at certain wavelengths. Also, in the final figure, the high TM0 in port 1 is surprising.

Larger suggestions (some of which may also help investigating accuracy issue):

- In your version of Fig. 2 (a), I notice that the TE0 mode is showing a small TE fraction at the smaller wavelengths. I'm not sure why this would be the case.
- Could you please use EMELengthSweep instead of running a batch of simulations, even for the taper length optimization? Scale factors which vary by cell_index are now supported.
- It would be better to use EMECompositeGrid consisting of EMEUniformGrid subgrids for the composite taper structure, rather than a single EMEUniformGrid.
- Can you check if the spikes which appear in the Y-junction EME simulation disappear if the number of modes is increased?
- Lengths of first and third tapers do have an effect, I suppose. you just fix them and then optimize the length of the second taper. Perhaps if you use a different length of first and third taper, you could see a better conversion efficiency?

Minor comments:

- In the mode solving step, can you note that we are using the local mode solver, which may result in lower accuracy because it does not use subpixel? Also, make a note of how long (30 s) these cells may take to run.
- I see lots of comments on how FDTD is more accurate than EME. This is true, but it is a bit subtle. Both can be quite accurate as long as they are configured correctly, but both can give misleading results if misconfigured. FDTD is certainly more robust than EME. And it is good to check using multiple simulation methods.
- Can you give a bit of explanation about smatrix_in_basis when you use it? Or link to the notebook where @tomflexcompute explains the usage of this function.


Thanks again, I think this example is really great!

caseyflex:
I played with this a bit more. The 2D length sweep seems to work fine. It seems like the incomplete mode conversion is due to the details of the mode dispersion coming from the mode solver. Decreasing W1 and increasing W2 a bit so they cover the entire hybridization region leads to complete mode conversion. I'll send you directly the notebook I was playing with.

The spike in the Y-junction length sweep sometimes comes from not having enough EME cells. But you could also try a different value of num_modes. Is 6 modes appropriate?


A few more comments:

- I don't think we should suppress warnings. Maybe just acknowledge that the warning about not having a source is OK here.
- For the taper, the EME output port is also the physical output port, so I don't think we need the extra mode solver monitor and smatrix_in_basis. `sim_data.smatrix` should be enough here, right?
- You only really need num_modes=2 for the taper I believe. And min_steps_per_wvl=30 should be fine too.

caseyflex:
In fact, I wonder if the mode dispersion would be more accurate using the
remote mode solver. You could actually fairly easily use an EME simulation
together with an EMEModeSolverMonitor with 0 thickness in the transverse
direction to obtain effective index of each mode as a function of the
width. This would be fast and use the remote mode solver for better
accuracy.

alec-flexcompute:
Quick update:
By scaling specific cells, I'm getting a (jagged looking) match with the figure in the paper.
image
For this reason I think their claim that the lengths of the first and third tapers being irrelevant is incorrect.

Next I'll be looking to fix this jagged issue that's been showing up in the figures

tomflexcompute (approved):
Thanks @alec-flexcompute . The current mode solving is done via a loop. Since we support batch for mode solvers now, could you modify your script to run batch instead as demonstrated [here](https://www.flexcompute.com/tidy3d/examples/notebooks/BatchModeSolver/)? Should be a quick fix and it can speed things up quite a bit.

prashkh:
@alec-flexcompute Just took a look and the results look good now. Congrats on putting in all the hard work to get this out.

prashkh:
@caseyflex sorry I accidentally clicked merged since that showed as a default option for me.  We still need to get your feedback. I can revert the develop branch to the state before this merge and push it back. Let me know what you think.

alec-flexcompute:
@tomflexcompute Mode batch solving implemented! Thanks for the suggestion.

caseyflex:
Thanks! A few more comments:

- I see a note to Tom in cell 9, should get rid of that
- When you say "These results are fairly different from what is inside the paper", we should make a note that the discrepancy may be due to different dispersion relation due to subpixel averaging
- In cell 21, I think you are plotting the wrong thing. You seem to have used `l2_test_pts` again, so we are just seeing the middle taper have length 1um
- Can you extend the structure to get rid of the warning in cell [22]?
- Cell [25] has something hardcoded
- I see a solver warning in cell [24]. Could you check the log and see if it's anything important, or just the extend structure warning? (the error message about ArrayLike dimensions is my fault and will be patched hopefully in 2.7.2, although it obviously doesn't prevent it from working)

alec-flexcompute:
Thanks for the excellent catches Casey! They are implemented and the previous merge was reverted. The latest "added LNOI..." PR had the edits and is now merged

PR #129, opened 2024-06-26
Added installation instructions on external libraries

tomflexcompute:
Added instructions on pip installing external libraries. These are the notebooks using external libraries I found so far.


e-g-melo (approved):
Thanks @tomflexcompute! 

It seems very good to me!

PR #130, opened 2024-06-26
Added batch run mode solver tutorial

tomflexcompute:
This PR includes a dedicated tutorial on batch mode solver. The difference in batch run for FDTD and mode solver (in FDTD one first creates a dictionary of simulations and then a Batch vs in mode solver one creates a list of mode solvers and then uses the web module to run_batch) has caused some confusion among users. It also seems like `run_batch()` method is not documented in the API reference so we might need to fix that later. 

FYI @FilipeFcp this is a typical PR to submit a new tutorial.


e-g-melo (approved):
Hi @tomflexcompute! This is a very useful notebook. I have a notebook on waveguide dispersion as well. I will introduce the batch to conclude it.

I have only one suggestion: using "Mode Solver" instead of folder_name="default" (which also works, so that is not critical).

tomflexcompute:
> Hi @tomflexcompute! This is a very useful notebook. I have a notebook on waveguide dispersion as well. I will introduce the batch to conclude it.
> 
> I have only one suggestion: using "Mode Solver" instead of folder_name="default" (which also works, so that is not critical).

Thanks Emerson. Is your notebook very similar to this or is it a different topic?

e-g-melo:
> > Hi @tomflexcompute! This is a very useful notebook. I have a notebook on waveguide dispersion as well. I will introduce the batch to conclude it.
> > I have only one suggestion: using "Mode Solver" instead of folder_name="default" (which also works, so that is not critical).
> 
> Thanks Emerson. Is your notebook very similar to this or is it a different topic?

It is different. I am reproducing the results of a paper and showing how to calculate D and GVD

prashkh:
Notebook looks great and the example showing mode crossing for TFLN is fantastic!

Here is a minor comment:
Since the color scheme has white for 0.5 polarization value, it appears as if some points are missing. Maybe you could change it to a slightly different color so all points are visible?


Screenshot 2024-06-27 at 12 00 35 PM


Ready to go otherwise!

tomflexcompute:
Thanks @e-g-melo @prashkh !

PR #131, opened 2024-06-28
Added tapered waveguide dispersion calculation.

e-g-melo:
This example shows how to calculate dispersion parameters (D and GVD) using Tidy3D. I've introduced the `run_batch` function to run the mode calculation in parallel. Afterward, I will create a GUI version.


tomflexcompute (approved):
Thanks @e-g-melo for the great demonstration. It's a very useful demo that addresses an important problem in PIC. A few minor feedback and questions below:

- "Photonic integrated waveguides" -> "Integrated photonic waveguides"?

- In cell[6] the `build_mode` function returns a ModeSolver object but it's indicated to return a td.ModeData? Is that a typo?

- "a Mode Solver list" -> "a list of ModeSolver instances"?

- Consider linking `PolySlab` in the markdown comment to the API reference.

- In the section where you tested different values of `group_index_step`, could you make a quick comment on what the number means physically? According to the API reference, it "control the computation of the group index alongside the effective index. If set to a positive value, it sets the fractional frequency step used in the numerical differentiation of the effective index to compute the group index." It also seems like a larger group_index_step gives a more accurate result, which sounds a little counterintuitive for me. Isn't smaller step better?

e-g-melo:
> * In the section where you tested different values of `group_index_step`, could you make a quick comment on what the number means physically? According to the API reference, it "control the computation of the group index alongside the effective index. If set to a positive value, it sets the fractional frequency step used in the numerical differentiation of the effective index to compute the group index." It also seems like a larger group_index_step gives a more accurate result, which sounds a little counterintuitive for me. Isn't smaller step better?

Thanks for your comments @tomflexcompute!

It seems that two other sampling points are included for each frequency point to help calculate the mode group index. The `group_index_step` controls how far or close these additional frequency points are from the central frequency. So, let's suppose we have only 1 frequency point `freq`. When calculating dispersion, two additional points will be included at `freq - freq*f_stepgroup_index_step` and `freq + freq*f_stepgroup_index_step`.

Maybe @lucas-flexcompute could confirm, but it seems there is no physical meaning, it only controls the numerical differentiation.  Maybe if we have enough frequency points and they are very close to each other this parameter would not be necessary? I wonder if we could calculate a good initial guess for it.

alec-flexcompute:
Excellent job @e-g-melo! Just a couple very small suggests:
1. Could you add a bit more of a description of `group_index_step`? The current one, while concise, I think could be clearer if it were longer, like what you wrote in the above comment
2. Maybe a bit more description of what dispersion in the waveguide is/results in could be helpful
3. In your first cell, change "the Tidy3D's ModeSolver" to "Tidy3D's ModeSolver"

alec-flexcompute (approved):

lucas-flexcompute:
> Maybe @lucas-flexcompute could confirm, but it seems there is no physical meaning, it only controls the numerical differentiation. Maybe if we have enough frequency points and they are very close to each other this parameter would not be necessary? I wonder if we could calculate a good initial guess for it.

That's right, it's only for numerical differentiation. We can't control the user's input to know ahead of time when those will be required or not. Also, the exact formulation (adding a fraction of the frequency instead of a constant delta) plays a part into the calculation. User-provided points would fall on an arbitrary grid and the numerical differentiation formulation would be different.

e-g-melo:
Thanks @alec-flexcompute, @tomflexcompute, and @lucas-flexcompute! I've addressed all your comments here.

PR #133, opened 2024-07-05
Filipe/metasurface

FilipeFcp:
This is an notebook reproducing a vortex metasurface experiment.

Any feedback is most welcome.


alec-flexcompute:
Hi Filipe, this looks like an awesome example! A couple suggestions, all to of which are about presentation:
1. Include a diagram/plot of the simulation/planned simulation that shows the cross-section. This would show the source, monitor, a bit of the structure/substrate, and mesh override structure. This would make it easier for new users to understand the setup
2. Below each new section, a brief description of the section, basically explaining why the section has its own section

FilipeFcp:
Thank you Alec.
I implemented your suggestions. Anything else, please let me know.

alec-flexcompute:
Looks pretty good! One thing though, can you add plt.show() at the end of cells where you're plotting things? This gets rid of the annotation right before the plot
image

Another thing to consider is to put the cross section diagram up top where you're describing the simulation, but this is a mild suggestion

e-g-melo (commented):
Hi @FilipeFcp! This is a very nice example! The comparison between the simulations and the experimental results is very exciting.

My comments are mostly related to formatting:

1. Use the same style of existing notebooks to cite that paper. See an example [here](https://www.flexcompute.com/tidy3d/examples/notebooks/NanostructuredBoronNitride/). You can also include a link to that paper.
2. You could also reduce the image size in the introduction.
3. I am curious about the phase discontinuity in steps of `pi/4pi/4`. Could it be simply `1/16`?
4. In the introduction, we use to include some links to other examples. See this [example](https://www.flexcompute.com/tidy3d/examples/notebooks/NanostructuredBoronNitride/). That can help us get a better ranking on Google searches.
5. You have mentioned the variable `NN`, but it was never used. Would it refer to `N`?
6. "of a hemi-infinite silicon" -> "of a semi-infinite silicon"
7. I suggest using a space after the "#" in the comments and also using a consistent pattern throughout the text. I usually start them with a capital letter and finish them with a period. But Tom and others usually start them with a lowercase letter and don't use a period.
8. "Free parameters" sounds like something related to optimization parameters. Maybe you could simply remove that header and refer to them as "simulation parameters" throughout the text. 
9. Use [jblack](https://pypi.org/project/jblack/) to format your notebook. It helps us to create a uniform formatting style.
10. Cite variables using `. For example: "delta" -> `delta`
11. Before cell [9], it would be nice if you can include a reference to `Transformed` or a notebook explaining those transformations.
12. Include a `plt.show()` after after your plots to hide the text output. For example, in cell [12]. We also have a pattern for the session headers. The first letter of the notebook title should be in capital letters, and all the first letters of the secondary header words should be in capital letters.
13. For the variables' definition, I recommend using only lowercase letters and underscores. For example, you could use `source_time` instead of `sourceTime`.
14. In cells [13] to [18], I suggest you include the comments as markdown to break that long sequence of code cells.
15. I think it would be nice if you reference the paper figures explicitly. For example Figure 5(c), Figure 5(d). You could also adjust the sizes of the experimental figures to match those of the simulation ones.
16. Before cell [58] it would be interesting to elaborate more about the differences between the field projections. Which one is more close to the expected result? Why?

tomflexcompute:
Thanks @e-g-melo for the great suggestions! @FilipeFcp let me know once you have incorporated Emerson's comments. I'll take a closer look at the notebook after that to avoid making redundant comments.

FilipeFcp:
Thanks @e-g-melo .  I believe I have already implemented everything. Right now, I am working on creating a better diagram and image of the model. However, I've noticed that the structures are not appearing in the 3D plot, neither in the notebook nor on the Tidy3D site.

e-g-melo:
> Thanks @e-g-melo . I believe I have already implemented everything. Right now, I am working on creating a better diagram and image of the model. However, I've noticed that the structures are not appearing in the 3D plot, neither in the notebook nor on the Tidy3D site.

Hi @FilipeFcp! Thanks for addressing the comments. The notebook looks very nice! 

Currently, GUI doesn't support visualization of GeometryGroup containing Transformed objects. I will see when we can implement that.

FilipeFcp:
Alright. Thank you once again, @e-g-melo, for the careful review and for taking the time to educate me about our notebook standards.

tomflexcompute (changes requested):
Thanks @FilipeFcp for the great notebook. It fills in another gap our current example library has. A few formatting comments below:

1. Consider adding a more general introduction about vortex beam generation using metasurface at the very beginning. For example, something like

"Vortex beam generation using metasurfaces leverages ultra-thin, planar structures to impart helical phase fronts to light, creating beams with orbital angular momentum (OAM). These metasurfaces consist of carefully designed arrays of nanostructures that manipulate light at subwavelength scales, enabling the creation of vortex beams with high efficiency and precision. By engineering the phase profile of the metasurface, vortex beams with specific topological charges can be produced, offering compact and versatile solutions for applications in optical communication, quantum information processing, and particle manipulation."

2. In general we might want to avoid using images from publications since the copyright of the images belongs to the publisher, which is AAAS in this case. We can replace them with images we draw such as the one below. For the comparison to experimental data in the end, we can also remove the images but just leave comments pointing the reader to see the original paper. 

3. To align with our naming convention, the notebook name can be changed to "VortexMetasurface.ipynb".

4. Only the title of the notebook can be h1 header. All subsequent headers need to be h2 or lower, e.g. "## Simulation Setup". Otherwise the documentaion page will not be built correctly.

5. There is a warning about the distance to PML being smaller than half a wavelength. Clearly the setup works fine but just for presentation, maybe just add a little more space to get rid of the warning?

6. Since plot_3d() doesn't show the correct geometries, we can remove it.

7. web.estimate_cost() will automatically print the cost now so we don't need the print() statement since it will print out an additional number. 

8. For educational purpose, we encourage showing the keyword arguments. 

results = web.run(
    sim,
    "vortexMetasurface",
    folder_name="data",
    path="data/%s.hdf5" % "1",
    verbose="True",
)

to 

results = web.run(
    simulation=sim,
    task_name="vortexMetasurface",
    folder_name="data",
    path="data/%s.hdf5" % "1",
    verbose="True",
)

9. The name of the notebook needs to be added to `docs/case_studies/metamaterials_gratings_periodic.rst` so the notebook will be displayed when the docs are rebuilt. 

10. For search engine optimization (SEO), we need to add metada to the notebook file. You can follow this [tutorial](https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e) and other existing notebooks on how to do so.

FilipeFcp:
I have updated the files @tomflexcompute. I believe I have addressed everything. I used your text with minor modifications, as it was already very good. I changed the simulation to avoid the warnings, but it is now costing 4.2 FC instead of 1.8.

e-g-melo (approved):
Thanks @FilipeFcp!

tomflexcompute:
Thanks @FilipeFcp for the revision. Looks very good. Just last two minor things:

1. There is an extra space between the DOI and the "." in the end.
image

2. The size of the image is a bit large. Usually we set the width to 400 pixels. You can take a look at other notebooks for example. 

After fixing them, you could squash all commits and then rebase and merge.

PR #134, opened 2024-07-08
Update autograd notebooks to use `get_intensity()` where appropriate

yaugenst-flex:
Update notebooks to use new differentiable `interp` functionality provided by https://github.com/flexcompute/tidy3d/pull/1769


tylerflex (commented):
Looks good, thanks!

just checking, should we re-run the antenna one to be safe? or is it fine / did you already re-run it?

yaugenst:
I fully ran the metalens one, the antenna one I only ran for a few iterations. Didn‘t notice any differences though so I thought that‘d be ok?

tylerflex:
ok, sounds fine to me then, thanks!

PR #141, opened 2024-07-19
Clarifying polarization

FilipeFcp:
Just a small clarification on the polarization definition. One user pointed out that since the waveguide mode is propagating in the x-direction and Ex=0, it should be a TE mode. The notebook calls it a TM mode, which I believe is the convention for the whispering-gallery mode resonance in the ring (with Hz parallel to the ring axis).

I just highlighted that it refers to the resonator mode.


tomflexcompute (commented):
Sorry @FilipeFcp I don't see any changes actually. What exactly is changed?

FilipeFcp:
@tomflexcompute Before cell [8]. I just added the word resonator:

'mode_index=0 corresponds to exciting 0-th order TM **resonator** mode (E=Ez) '

I thought that it would be enough.

tomflexcompute:
> @tomflexcompute Before cell [8]. I just added the word resonator:
> 
> 'mode_index=0 corresponds to exciting 0-th order TM **resonator** mode (E=Ez) '
> 
> I thought that it would be enough.

Oh sorry I missed that. Maybe it's worthwhile to elaborate a bit on the mode definition to avoid confusion.

In the printed mode properties, it actually indicates TE fraction of 1 for the first mode and TM fraction of 1 for the second mode, which can also be confusing to see the conflict. I wonder if we should stick to this convention as it's more commonly accepted.

FilipeFcp:
> > @tomflexcompute Before cell [8]. I just added the word resonator:
> > 'mode_index=0 corresponds to exciting 0-th order TM **resonator** mode (E=Ez) '
> > I thought that it would be enough.
> 
> Oh sorry I missed that. Maybe it's worthwhile to elaborate a bit on the mode definition to avoid confusion.
> 
> In the printed mode properties, it actually indicates TE fraction of 1 for the first mode and TM fraction of 1 for the second mode, which can also be confusing to see the conflict. I wonder if we should stick to this convention as it's more commonly accepted.

You are right. I believe most people will expect the waveguide definition, and those familiar with ring resonators will understand what is happening.

I updated the code to follow the waveguide definition and reran everything to ensure the last plot reflects these changes. I only modified markdown before cell [8] and cell [16].

tomflexcompute (approved):
Looks good. Thanks @FilipeFcp .

alec-flexcompute:
Ah, this is a good catch, and shows I need to study up on my mode definitions as well! I think the updated note is succinctly put

alec-flexcompute (approved):

tomflexcompute:
Thanks @alec-flexcompute . Indeed TE and TM definitions can be confusing depending on the system and dimensionality. 

@FilipeFcp please squash the commits before merging.

FilipeFcp:
Thanks everyone!

PR #142, opened 2024-07-24
broadband adjoint notebooks

tylerflex:


e-g-melo (approved):
Thanks @tylerflex! You provided another great example of adjoint!

### Autograd17BandPassFilter.ipynb

1. I suggest you initialize the parameters in cell [13] using random values to make the images in [13] and [11] different from each other.
2. I wonder if you could use all those stuffs in `tidy3d.plugins.invdes` like `FilterProject`, `TopologyDesignRegion`, `ErosionDilationPenalty`, etc. in this notebook.
3. In [27], maybe use `sim_final.plot_eps(z=0, monitor_alpha=0)` to hide the monitor.
4. Would you obtain the same result if you set symmetry about the `y`-axis?
5. `weighted_sum` is defined but never used.
6. Is there a specific reason for choosing that normalization between -1 and 1 in `get_metric`? I wonder if you would obtain a similar result by minimizing the RMSE between `target` and `transmission`.

### Autograd9WDM.ipynb
1. Could you change `Base Simulation` to `Static Components` as you did in the band-pass filter? I will use `Static Components` in GUI, so it would be better if we start to unify the naming.

Autograd17BandPassFilter.ipynb
yaugenst-flex, line 1: Agree with @e-g-melo about symmetry, I think that could be included here, and just from a physical consideration I think this is known a priori.

[2]:
 - Where does this `assert num_freqs % 2 == 1, "num_freqs must be odd"` condition come from?

[16]:
 - `zeros = np.zeros_like(freq)` not used
 - `1.0 * (np.logical_and(wvl > wvl_min, wvl < wvl_max))`: can just return `(wvl > wvl_min) & (wvl < wvl_max)`, the casting isn't necessary

[17]:
 - `weighted_sum = np.sum(transmission * weights)` unused
 - This is maybe a matter of preference, but if you don't cast the "weights" in `targets()` to float, you can just use them as an index. Of course they should be called something like `mask` or such instead. So:
     ```python
    def get_metric(transmission: np.ndarray) -> float:
        mask = target(freqs)
        sum_measured = np.sum(transmission[mask])
        sum_best = np.sum(mask)
        sum_worst = np.sum(~mask)
    
        value = (sum_measured - sum_worst) / (sum_best - sum_worst)
        
        return value
    ```
    Actually, now that I'm writing this down and following the logic, I'm not sure what `sum_worst` and `sum_best` are supposed to be? They are independent of the transmission, and are just the no. of points inside vs. outside of the bandwidth?

Autograd9WDM.ipynb
yaugenst-flex, line 1: [Setup]:
 - pixellated -> pixelated (appears somewhere else too I think, ctrl+f)
 - input light of different wavelengths get**s**
 - adding dimension to the structures -> adding _a third_ dimension?

[6]:
 - `funciton` -> `function`

[8]:
 - Minor, but would `np.random.uniform(0.4, 0.6, (nx, ny))` potentially lead to a bit better results?

[18]:
 - Is the output excluded intentionally?

Between [18] and [19]:
 - Opimization -> Optimization
 - We will make use **of** the

[25]:
 - Is there a reason why we optimize for mode power but plot the flux?
 - Remove comments

[46 ff.]:
 - I saw that animation, I think `fps=60` is too high, really hard to see anything.

yaugenst-flex (approved):
Looks great overall! Maybe run both notebooks through `ruff format` once? Usage of space and single vs. double quotes is a bit inconsistent. And maybe even `ruff check`, because there might be some unused variables that I didn't catch.

Autograd9WDM.ipynb
tylerflex: > [18] Is the output excluded intentionally?

No I just added the print statements after running so they'd show up next time.

> [25] Is there a reason why we optimize for mode power but plot the flux?

Not really. I just figured it might be nice in some sense to just plot the total power flux. Should I plot the power only in the fundamental mode?

Made all other changes, thanks.

yaugenst-flex: > Not really. I just figured it might be nice in some sense to just plot the total power flux. Should I plot the power only in the fundamental mode?

At the very least I'd be curious how different the two are. Assuming that the waveguides are single-mode, they should be pretty similar, but the flux will be underestimating the loss in the device slightly. Maybe just try and see how it looks?

Autograd17BandPassFilter.ipynb
tylerflex: > Where does this assert num_freqs % 2 == 1, "num_freqs must be odd" condition come from?

I wanted there to be a frequency component at freq0, but I think it's not necessary now as I can interpolate.

> Actually, now that I'm writing this down and following the logic, I'm not sure what sum_worst and sum_best are supposed to be? They are independent of the transmission, and are just the no. of points inside vs. outside of the bandwidth?

Yea I dont really like my objective function anymore. I changed it to this:

```py
def get_metric(transmission: np.ndarray) -> float:

    in_band = target(freqs)

    avg_transmission = np.mean(transmission[in_band])
    avg_supression = np.mean(transmission[~in_band])
    
    # ideally,  (avg_transmission = 1 and avg_supression = 0) -> metric = 1
    # at worst, (avg_transmission = 0 and avg_supression = 1) -> metric = -1

    return avg_transmission - avg_supression

```

we'll see how that works

yaugenst-flex: Yeah that makes intuitive sense to me!

tylerflex:
Thanks @e-g-melo , I made all the changes.

> Is there a specific reason for choosing that normalization between -1 and 1 in get_metric? I wonder if you would obtain a similar result by minimizing the RMSE between target and transmission.

I simply like normalizing the objective function contributions because I get a more intuitive feel for how well the optimization is performing and also it makes the gradient unitless (if the parameters are unitless), which can help with setting a proper learning rate.

Symmetry would be -1 in y?

I made an analogue to this notebook using the inverse design tool. But I'm not sure which I like more. Maybe I'll include a short section at the end showing how it could be set up, but not run it more than a few iterations?

e-g-melo:
> Symmetry would be -1 in y?

Yes

e-g-melo:
> I made an analogue to this notebook using the inverse design tool. But I'm not sure which I like more. Maybe I'll include a short section at the end showing how it could be set up, but not run it more than a few iterations?

It sounds very good! This way, the user can choose the route they prefer.

tylerflex:
Thanks all changes have been made!

PR #143, opened 2024-07-24
uploading the notebook

FilipeFcp:
Hi @tomflexcompute. I am opening the PR.

I approached the medium fitting slightly differently. Instead of downloading the data, I integrated both the download and processing steps into the Python code. I think this method can be transferable for other datasets from [refractive.info.](https://refractiveindex.info/)

I need to create the image. The one that is there is just temporary. I think I will try to use Blender's Python interface for this.


tomflexcompute (changes requested):
Thanks @FilipeFcp . Result matches the experimental data perfectly. Myth busted! This example works as a great social media story.

A few rough corners in formatting:

1. Title (H1 header) is missing. 

2. The citation format seems slightly different. Usually we list the authors as well. 

3. The comment "The material properties of the TiO₂ are obtained directly from a website and fitted using Tidy3D's dispersion fitting tool." seems redundant since it's mentioned in the previous paragraph already.

4. "were you can find..." -> "where you can find..."

5. "case studies such as the scattering cross-section calculation of a dielectric sphere, the scattering of a plasmonic nanoparticle , the gradient metasurface reflector, and the dielectric metasurface absorber."

6. Since the introduction is already somewhat long, I would skip the "If you are new to FDTD..." and "FDTD simulations can diverge..." paragraphs in this case.

7. "Getting optical properties data and creating the medium" -> "Retrieving refractive index data and creating the medium".

8. "...losses of Ti02 are in..." -> "...losses of TiO$_2$ are in..."

9. "can be installed via `pip install pyaml`.

10. "building de simulation" -> "building the simulation"

11. Is there a particular reason why you created a `monitors` list and appended two monitors to it instead of having monitors=[monitor1, monitor2] in initializing a Simulation object?

12. In the plots, add x label to indicate x axis is wavelength in microns.

FilipeFcp:
Thanks, @tomflexcompute. Sorry about all the silly mistakes. Thank you for the careful review.

tomflexcompute:
Thanks Filipe. Just realized I forgot to point out that you need to add the notebook file name to docs/case_studies
/nanophotonics.rst or another category that you think is more suitable.

FilipeFcp:
> nanophotonics.rst

I forgot about that. I just did it, thanks.

e-g-melo (approved):
Thanks for this great example @FilipeFcp!

I have only a few comments:
1. "Metal oxied sunscreen" -> "Metal oxide sunscreen"
2. "create dataframe with data with the data" -> "create a dataframe with the data"
3. I suggest reducing the size of the introductory image a bit.
4. It would be nice if you include a legend entry for all the curves in [4]. Something like in the code below:
```py
source.source_time.plot(np.linspace(0, t_apod, 2000), ax=ax)
apodization.plot(times=np.linspace(0, t_apod, 2000), ax=ax)
lin = ax.get_lines()
lin[0].set_color("red")
lin[1].set_color("blue")
lin[0].set_label("source")
lin[1].set_label("apodization")
plt.legend()
```
5. "We are simulate for two active" -> "We will simulate for two active"
6. It would be nice to tell the readers at which wavelength they should pay attemption in the final plots. For example: "Looking at the proportion of reflected light below XXX wavelength..."

Best,
Emerson

alec-flexcompute (approved):
Hi Felipe, excellent job! A couple suggestions:
1. Above cell 2, 'dispertion' should be 'dispersion'
2. Is there a reason why the sim size in z is that large? I think you may be able to decrease its size
3. Maybe a short note on why your source is a good model for sunlight

FilipeFcp:
Thank you @e-g-melo  and @alec-flexcompute  for your comments. I took some time to update the notebook as I wanted to learn Blender to produce the images. I have implemented your corrections and suggestions and added a new image.

tomflexcompute:
Thanks @FilipeFcp . Figure looks nice. I just went over the notebook again and ran a pass of Grammarly. A few typos were caught. After fixing them, it should be good to go.

- "larger than 5 FlexCredit" -> "than 5 FlexCredits". In fact, we usually only say it's larger than 1 or larger than 10 credits since the cost formula might change slightly in the future. In this case, we can just say more than 1 credit. 

- "developed since the 1940s" -> "developed in the 1940s "

- "we will reproduce results presented" -> "we will reproduce the results presented"

- "We will retrive material" -> "We will retrieve material"

- "material propertie data" -> "material property data"

- "Defing source and monitors" -> "Define source and monitors"

- "two active particles proportions" -> "two active particle proportions". Instead of "proportion", I wonder if "concentration" is more suitable? 

- "in pararell using" -> "in parallel using"

@weiliangjin2021 In this example Filipe had to manually pull data from refractiveindex.io using urllib. I remember a while back our built-in `from_url` function had some issues but that has been fixed?

weiliangjin2021:
>  In this example Filipe had to manually pull data from refractiveindex.io using urllib. I remember a while back our built-in from_url function had some issues but that has been fixed?

I think `from_url` has been working fine, but we have some outdated URLs in our examples and docs.

tomflexcompute:
> I think `from_url` has been working fine, but we have some outdated URLs in our examples and docs.

Thanks. @FilipeFcp could you give it another try and get help from Weiliang if needed?

FilipeFcp:
Thanks @tomflexcompute. I installed a grammar revision extension and am testing Grammarly so that these errors won't happen anymore.

@weiliangjin2021 It works fine indeed, but I could not make it work for YAML files. Is that implemented?

FilipeFcp:
@tomflexcompute I changed the notebook, so it is using the `from_url` method.

tomflexcompute (approved):

tomflexcompute:
> @tomflexcompute I changed the notebook, so it is using the `from_url` method.

Thanks @FilipeFcp looks great. I think it's ready for squashing and merging.

PR #147, opened 2024-07-29
add CircularlyPolarizedPatchAntenna

dmarek-flex:
Add an RF antenna notebook. I am not using the Directivity monitor at the moment because I need to decompose the field in to RHCP and LHCP, which I don't think is quite ready.


dmarek-flex:
@QimingFlex Any ideas on how to extend the DirectivityMonitor to perform some of the computations needed in this notebook? For these results, I need to isolate radiated fields based on their polarization.

QimingFlex:
Yeah, how about introducing two more fields to the DirectivityMonitor?

dmarek-flex:
> Yeah, how about introducing two more fields to the DirectivityMonitor?

That would work, I just foresee us needing to add new types of results in the future that are all fairly simple computations.

We could create the RadiationMonitor which records radiation intensity U and is decomposed into the different polarization states (linear-> Ephi and Etheta, circular->RH and LH). Then, if we also record the total flux we can compute the Directivity in and other derived quantities as `@cached_property`s.

Or even more flexible, I wonder if it would make sense for the RadiationMonitor to simply store the electric fields, and the remaining quantities are computed when desired  as `@cached_property`s.

QimingFlex:
Oh, actually we can put RH and LH as `@cached_property` to the far field monitor instead of putting under the directivity monitor.

dmarek-flex:
Right now I cannot make use of the DirectivityMonitor because of the specific values I need to compute. In the future, we can replace many of my operations with an updated DirectivityMonitor.

tomflexcompute (approved):
Thanks @dmarek-flex for another great antenna example. The workflow is very nice. Manually adding the override structures is a bit tedious but hopefully it can be automated in the future. Some minor issues I found:

- "as snapping points to ensure that their their precise positions" -> remove one "their".

- "Using a smaller cell size would lead to a more accurate result, but since the FDTD time-step size is linked to the smallest grid cell size through the Courant condition. Therefore, making the cell size smaller can significantly increase the duration of the simulation." -> These two sentences read a bit strange. Maybe "Using a smaller cell size would lead to a more accurate result, but since the FDTD time-step size is linked to the smallest grid cell size through the Courant condition, making the cell size smaller can significantly increase the duration of the simulation."?

- The last three plots have label="tidy3d". Did you mean to plot the paper results as a comparison? There could be some copyright complications if we directly use the data from the paper.

dmarek-flex:
Thanks for finding those issues. Will you be generating a figure for this notebook as well?

> * The last three plots have label="tidy3d". Did you mean to plot the paper results as a comparison? There could be some copyright complications if we directly use the data from the paper.

Oh yea good find! I have some digitized results from the paper that we will use for testing internally. But I wasn't planning on including any other results in this notebook. I'll remove the label.

tomflexcompute:
Yes I'll make an image tomorrow morning.

tomflexcompute:
@dmarek-flex could you also add metadata to the notebook? Otherwise the learning center rebuild won't include the thumbnail image correctly.

weiliangjin2021:
Great notebook!

- "Fortunately, we can use the [FastDispersionFitter](https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.plugins.dispersion.FastDispersionFitter.html#tidy3d.plugins.dispersion.FastDispersionFitter) to easily generate such a model": might be more helpful by directing to this convenient function https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.plugins.dispersion.FastDispersionFitter.html#tidy3d.plugins.dispersion.FastDispersionFitter.constant_loss_tangent_model
- Once @QimingFlex's PR is merged, the gain and the axial ratio can be computed with DirectivityMonitor?

QimingFlex:
Strong notebook in patch antenna @dmarek-flex ! 
I think we can move RHCP and LHCP components to `FieldProjectionAngleData ` as cached properties in the future?

weiliangjin2021:
Since Damian if on paternity leave, @QimingFlex could you take over and also get familiarized with doc?

tomflexcompute:
@QimingFlex if you are taking over this PR while Damian is away and in case you are not familiar with it, you can see how and what to add in the metadata [here](https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e).

weiliangjin2021 (approved):

PR #149, opened 2024-08-01
Nanobeam notebook

FilipeFcp:
Hi everyone,

I’m opening the pull request for the nanobeam example. It might be a bit lengthy at this stage. If you could take a quick look and let me know if you think all the results are relevant, I would appreciate it. I tried to make it generic, so it can be used for different geometries.

If you find it is too much information or too complex, I can simplify it.


alec-flexcompute (approved):
Hi Felipe, great notebook! I have a couple suggestions:
1. After reading this notebook through, I'm still a bit unclear as to the physics of this mode confinement. I think in the intro, it might help to describe more of the calculations you'll be making, and why they are relevant to the device
2. In the taper polynomial, this TeX rendering has a backslash here: 
image
3. Above cell [5], "refers to the total number tapered unit cells" should have an 'of' in it :)
4. When noting the calculation of mode volume above cell [18], maybe print out the calculated 0.39(\lambda / n)^3 value next to the mode volume value that's printed
5. Above cell [19], maybe describe the directional Q factor phenomenon here a bit more as it relates to the device? As someone who is unfamiliar with these types of devices, it might be helpful to supplement this.

tomflexcompute (commented):
Thanks @FilipeFcp for another content-rich notebook. Great demonstrations of many things that users would like to see. Some formatting comments:

- "DOI: https://doi.org/10.1063/1.3107263)." -> there seems to be an extra ")" in the end. 

- "tapered parameter" -> maybe "taper parameter"? Not totally sure.

- In our convention, in h2 headers, the first letter of every word is capitalized.  

- "that will accumulate the fields in different positions of the cavity" -> "that will record the fields as a function of time in different positions of the cavity"

- The function naming convention is a bit inconsistent. `taperFactorCubic`, `AssembleUnitCells`, `ellipse_uc`, etc. Is there a specific meaning associated with each way of naming? 

- In cell[12], I wonder why the number 0.1002200045334297 got printed out. It shouldn't happen I think?

- In cell[13] we see the warning that field decay value is large. Here we can make a comment to reiterate that large field decay is normal and expected in this case. 

- Above cell [14]: "where is possible to observe a tightly confined mode" -> "where it is possible to observe a tightly confined mode."

In cell[18], maybe print something like "Mode volume: 0.43 $(\frac{\lambda}{n})^2$".

e-g-melo (approved):
Hi @FilipeFcp! Thanks for this awesome example!

It looks very nice. I have only a few suggestions:

1. Define $\Delta t$ and $\Delta v$.
2. If "\#" means "number", I suggest writing "number" explicitly.

image

3. I think you could replace $Md$ with $M_d$
4. I wonder why you have set `start=0` on the point monitors and created that boolean mask instead of setting `start=start_time` directly.
5. I suggest including some comments on the limitations for the mode volume calculation.

Best,
Emerson

FilipeFcp:
Thank you all for the suggestions:

@alec-flexcompute: I added more information to the introduction and about the directional Q-factor. Please let me know if you think it is better or still needs further enhancement.

@tomflexcompute: I believe I addressed everything. I have also purchased the paper.

@e-g-melo: Regarding point 4, sometimes I find it useful to look at the signal and manually decide which part of the data to analyze, so I believe that recording the fields and applying the boolean mask is more flexible. Do you think it would be better to apply the start time directly in the monitor for this example?

e-g-melo:
> Do you think it would be better to apply the start time directly in the monitor for this example?

Got it! It would be interesting to include this information for the users so they can better understand your motivation.

tomflexcompute:
That's a great point. I think what Filipe demonstrated here is also very helpful. Maybe @FilipeFcp can add a comment on the reason for this and also say something like "Here we apply a filter because... . You also can set a start time at the monitor as we demonstrated in other examples"

FilipeFcp:
Thanks for the suggestions. I included some explanations while defining the sources and in the data analysis part.

tomflexcompute:
Sorry for some last minute change. Yesterday we started adding new fields in the metadata. See the last section [here](https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e). Could you add the "applications" and "features" fields to the metadata? In this case I think something like

"applications": ["Passive photonic integrated circuit components", "Photonic crystals", "Quantum emitters"],
"features": ["Resonance finer"],

FilipeFcp:
Done :)

tomflexcompute:
> Done :)

Thanks. Feel free to squash and merge.

PR #150, opened 2024-08-01
updating the frequency value

FilipeFcp:
Hi guys.

I ran this notebook, and the mode volume yields a different value: The example is 0.81, and running the notebook now yields 0.17. 

I was investigating, and the thing is that the resonant frequencies are slightly different now (the original was 3.291688e14, now is 3.297079e14). This is just 1.5 nm, but affects the field monitor since the Q-factor is high. 

I just ran again with the updated frequency value, so it yields a correct value for the mode volume.


alec-flexcompute (approved):
Looks good - good job finding the fix!

tomflexcompute (approved):
Thanks @FilipeFcp for spotting and fixing it. 

It seems like a smaller change in result due to some solver changes. Not sure if it's worrisome but FYI @momchil-flex .

momchil-flex:
Hmm, interesting. When it comes to the small change in frequency, it would be good to understand, but I think it may be generally fine since these high-Q cavities are great "sensors", i.e. they amplify every small change, so it really could have been something very tiny that has changed that is only visible in such a test.

What I'm surprised by is that this affected the mode volume computation so much. This part has a slightly tricky bit in that we seem to need to know the resonance frequency in advance to define the monitor, while in practice if that's the case we'd have to run two simulations of course. The thing is though that I have a feeling that I never had to do that when working with PhC cavities, i.e. my frequency monitors could be just *approximately* close to the resonance frequency and I would get the right values, because the fundamental mode is so strongly dominant that you always get bascially the same fields, apart from a scaling factor. The scaling factor is always irrelevant though since the normalization is always going to be arbitrary when using apodization (and not letting the simulation fields decay).

After looking into this, I think the problem is the apodization definition. I think whoever made this notebook did not fully understand the `width` argument, but in fairness the documentation may not be too clear, or we should also provide an image. This is not the width window in which fields are recorded; it is the width of the ramping up of the scaling function from 0 to 1, where it would be 1 at `t = start`. So the definition shouldn't be

```
apod = td.ApodizationSpec(start=t_start, width=run_time - t_start)
```

But more like, say

```
apod = td.ApodizationSpec(start=t_start, width=t_start / 10)
```

You can use e.g. `apod.plot(sim.tmesh)` to visualize the scaling function.

I think what was happening was that the source was being captured especially away from the resonance, which would explain the smaller mode volume computed (strong field in the center due to the source).

@FilipeFcp as a follow up could you:
1. Try to modify this notebook once again to fix the apodization spec, and see if you can provide only a rough estimate for the cavity resonance frequency (e.g. as taken from the paper) and still get the same result
2. Propose an update on our ApodizationSpec documentation to make things a bit more clear, maybe even adding a plot with labels to explain the start, stop, and width parameters

Thanks!

FilipeFcp:
Thank you @momchil-flex It was the apodization definition indeed. I tried with your suggestion, and the mode volume stayed the same for all the source spectrum (900 nm to 920 nm). I opened a PR for correcting the apodization.

tomflexcompute:
That's very nice!

PR #152, opened 2024-08-05
Filipe/cavity fom

FilipeFcp:
Fixing the apodization specification


momchil-flex:
Thanks. Sorry to keep digging into this, but since we should be pedagogical in our examples, I think we should either remove the `colocate=False` in the volume monitor, or, instead of using `simulation.epsilon_on_grid`, we should add a `PermittivityMonitor` with the same settings as the volume field monitor and use the epsilon from there. The latter is a bit easier - maybe also it will make the mode volume computation get closer to the paper value?

Will you also follow up on clarifying the apodization spec documentation?

FilipeFcp:
Hi @momchil-flex.

The mode volume using the `PermittivityMonitor` is 0.79, and using `simulation.epsilon_on_grid` is 0.81. 

I updated the notebook using the `PermittivityMonitor` as the value is closer to the paper.

We are working on the documentation. Change the text to the meaning of the width argument to be more explicit, and try to add an image.

momchil-flex (approved):

PR #154, opened 2024-08-06
Tom/adding tags

tomflexcompute:
After discussing documentation improvements with Emerson, the first step we can take is to add tags, addressing this [issue](https://github.com/flexcompute/tidy3d/issues/744). 

Also discussed with MC on formatting. The current approach adds two fields in the notebook metadata, a "applications" field specifying the area of the example like our current categories, and a "features" field specifying what features are used in the example. Both fields can take multiple tags. The Learning Center will then provide options to filter examples through tags. Hope this will make finding the right example for reference easier for users. 

I took a first pass in this PR but feel free to suggest any other tags we can include.


AndersonLocalization.ipynb
tylerflex, line 1: could maybe argue for this being a parameter scan feature demo, although it doesn't use any of the helpers as far as I know

BroadbandDirectionalCoupler.ipynb
tylerflex, line 1: should passive devices and integrated circuit components be separate applications?

tylerflex (approved):
looks good @tomflexcompute I'm also wondering if we want to add features "inverse design" or "adjoint optimization" etc?

tomflexcompute:
@tylerflex I think at one point we were thinking about moving some adjoint examples from tutorials to case studies but didn't end up doing it. Don't recall the reason exactly but we can discuss it again when you visit Boston later this week.

BroadbandDirectionalCoupler.ipynb
tomflexcompute: I think we mainly want to separate passive and active PIC components since one requires other solvers like heat and the other only requires FDTD.

tylerflex: sounds good

e-g-melo (approved):
Hi @tomflexcompute! Thanks for that. It will certainly help users.

I have only a few suggestions for additional applications and features.

1. Add "Structural Color" to `AllDielectricStructuralColor.ipynb` and `MIMResonator.ipynb` applications.
2. Add "Sensor" to `CMOSRGBSensor.ipynb` and `BiosensorGrating.ipynb` applications.
3. What about including "Quantum Emitter" to `BullseyeCavityPSO.ipynb`?
4. Add "Mode analysis" and "Parameter Sweep" to `FocusedApodGC.ipynb` features.
5. Add "Plasmonics" to `PlasmonicNanorodArray.ipynb`, `PlasmonicWaveguideCO2Sensor.ipynb`, `PlasmonicYagiUdaNanoantenna.ipynb`, and `MIMResonator.ipynb` applications.
6. Add "PerturbationMedium" to `ThermallyTunedRingResonator.ipynb` and `ThermoOpticDopedModulator.ipynb` features.

Best,
Emerson

PR #155, opened 2024-08-08
updating repo links

FilipeFcp:
I noticed that some other notebooks have outdated repo links.

It was: https://github.com/flexcompute-readthedocs/tidy3d-docs/tree/readthedocs/docs/source/notebooks/misc
and now is: https://github.com/flexcompute/tidy3d-notebooks/tree/develop/misc

I run a simple Python code to find and replace them. I checked, and seems the files needed for the notebook are indeed in the new link.

I am not sure @tomflexcompute if I should open a PR for each file, or if it is fine to do like this.

### DIFF
--- a/8ChannelDemultiplexer.ipynb
+++ b/8ChannelDemultiplexer.ipynb
@@ -3213,7 +3213,7 @@
    "source": [
     "Once we confirm the efficiency of the six couplers individually, we are ready to put them together to construct the entire (de)multiplexer device. This device is made by connecting the couplers via linear tapers. The linear tapers have a small angle of 1.8 degree to ensure good transmission. The overall device has a length of about 200 $\\mu m$.\n",
     "\n",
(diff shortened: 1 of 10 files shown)


tomflexcompute:
That's a great catch! Thanks @FilipeFcp . 

On a somewhat irrelevant note, could you fix some other typos and lump them into this PR so I don't have to open another one for them? The first one is in the genetic algorithm notebook. "We will an array" -> "We will use an array".
image

The other one is in the nanobeam notebook you merged yesterday. In the metadata, you have "Resonance finer" instead of "Resonance finder" in the "features" field.

alec-flexcompute (approved):
Nice catch @FilipeFcp , and good point @tomflexcompute! @FilipeFcp out of curiosity, could you share what you used to find these instances? Would be good to have this onhand in case there's future issues

FilipeFcp:
Done

tomflexcompute (approved):
That's great. Thanks @FilipeFcp . Feel free to squash and merge.

PR #156, opened 2024-08-09
adjusting fourier transform frequencies

FilipeFcp:
I believe the Fourier transform part of the  'bandstructure' notebook defines the frequencies in a nonideal way.
U modified using the built-in `fftfreq`. I think the new results better match the frequencies returned by `ResonanceFinder`.

image

image


caseyflex (approved):
Thanks for the catch! I agree, the previous frequency calculation didn't account for the fact that the monitor doesn't start recording at `t=0`.

While we're at it, can I suggest the following additional clean-up?

```
field = batch_data["sim_1"].monitor_data["monitor_time_0"].Hz.squeeze()

fmesh = np.fft.fftfreq(field.size, np.mean(np.diff(field.t)))
spectrum = np.fft.ifft(field)
mask = (fmesh > freq_range[0]) & (fmesh < freq_range[1])

plt.plot(
    fmesh[mask],
    np.abs(spectrum[mask]),
)
```

That way, my incorrect `df` calculation isn't needed. Also, `[::-1]` introduces an off-by-one error compared to `ifft` -- `ifft` should be correct here based on our conventions, I believe.

tomflexcompute (approved):
Great catch thanks @FilipeFcp !

FilipeFcp:
Sure, it looks better.
I just had to use field = batch_data["sim_1"].monitor_data["monitor_time_0"].Hz.squeeze().real,
to make sure that the resonances will appear in the positive frequencies.

e-g-melo (approved):
Hi @FilipeFcp! Thanks for looking into this!

Could you please take a look at this example as well: https://www.flexcompute.com/tidy3d/examples/notebooks/ResonanceFinder/?

Best,
Emerson

FilipeFcp:
@e-g-melo That one is correct

PR #157, opened 2024-08-12
Tagging inverse design example notebooks

tomflexcompute:
Tagging the inverse design example notebooks (the autograd ones specifically) so they consolidate with other examples in the learning center. The plasmonic antenna and bilayer grating coupler notebooks don't have a 3D schematic image yet so I skipped them in this PR but will find some other time to do them. 

Also adjusted some other example notebooks' tags according to feedback I received over the past few days.


e-g-melo (approved):
Thanks @tomflexcompute!

I wonder if we should also classify the features by "Topology Design" and "Shape Design". Would that add much complication?

tomflexcompute:
> Thanks @tomflexcompute!
> 
> I wonder if we should also classify the features by "Topology Design" and "Shape Design". Would that add much complication?

Yes I was also thinking about it. At the same time I think it would be nice to see all adjoint examples under one tag.

tomflexcompute:
Cover images and tags have also been added to the 3 latest example notebooks.

tylerflex:
Thanks @tomflexcompute, looks great. Agree with @e-g-melo that having more differentiation would be nice but for now a single tag is fine.  The images look great.

Do we want to leave the Adoint* examples untagged? or create a special tag like `Adjoint Plugin (Deprecated)`?

tylerflex (approved):

tomflexcompute:
> Thanks @tomflexcompute, looks great. Agree with @e-g-melo that having more differentiation would be nice but for now a single tag is fine. The images look great.
> 
> Do we want to leave the Adoint* examples untagged? or create a special tag like `Adjoint Plugin (Deprecated)`?

Yeah that's a good point. I wonder when we will fully transition into autograd and remove the adjoint plugin. Would it be 3.0? Not tagging the adjoint plugin notebooks has the advantage of making the autograd examples more visible so hopefully users, especially new users, are more likely to start using it over the adjoint plugin, which helps the transition.

tylerflex:
> I wonder when we will fully transition into autograd and remove the adjoint plugin. Would it be 3.0?

We are no longer adding new features to it, so any current users should probably be suggested to try the autograd plugin. But we'lll probably still support it in its current form until 3.0.

>  Not tagging the adjoint plugin notebooks has the advantage of making the autograd examples more visible so hopefully users, especially new users, are more likely to start using it over the adjoint plugin, which helps the transition.

Yea I think that's reasonable, although maybe with a tag like `Deprecated` in the title, it would be more clear. And I think we still want to be somewhat user friendly to adjoint users until maybe 2.9 when we can start gradually reducing the exposure on our docs?

tomflexcompute:
That sounds good. Shall we just change the title of the entire section in both the learning center and docs to something like "Inverse Design with the Adjoint Plugin (Deprecated)"?
image

tylerflex:
Let's leave the learning center alone I'd say,, but change the docs to "`Inverse Design`" (autograd) and "`Inverse Design with the Adjoint Plugin (Deprecated)`"

PR #158, opened 2024-08-14
Design Plugin Notebooks

m-bone:
New Bayesian optimization notebook and updated GA and PSO notebooks with the new Design plugin. Requires the Tidy3D branch [matt/bay_opt](https://github.com/flexcompute/tidy3d/tree/matt/bay_opt) to run.

Tom and Tyler would you mind taking a look and reviewing the code and text to check it reads well? Emerson I've added you in case you want other examples of the optimizers working.

ParticleSwarmOptimizationPBS has a better performance than previously; the GA and other PSO notebook have worse performance. The GA is because the initial state is being created differently and PyGAD calls the random state a different way. I cannot see the reason PSO is not repeating the previous example: the initial positions are identical and the same seed is being called the same number of times. I'm going to iterate with some other seeds and see if I can find a better example because the Purcell factor in the current one is poor.

Other Q's to answer before release:
- [x] Should we using early stop for the GA and PSO? Both have long plateaus at the end. It would demo early stop and it would reduce the notebook runtime
- [x] DesignSpace can easily set the local output directory for sims: should we use one? My "notebooks" directory is full of .hdf5
- [x] Why is `mesh_override` defined in cell 7 but never used? Can it be removed?
- [x] Cost warning at the start is "larger than 10 FlexCredits" in every notebook. Should we be more specific or is > 10 fine?
- [x] GA notebook needs to link to the BO notebook at the start once the BO notebook is on www.flexcompute.com/tidy3d/examples/notebooks/...


tomflexcompute:
It seems like our `MethodParticleSwarm` is using `pyswarm` under the hood? Even in this case, not sure if we need to expose it explicitly in the notebook?

tomflexcompute (dismissed):
Thanks @m-bone for the optimizer implementation and the demo example. It's very nice. 

For the Y branch Baysian optimization notebook:

1.  Could you expand a little bit in the comment on the specific form of the objective function? 
2.  After the optimization, we can add the same thing we did in the PSO notebook where we add field monitor and perphas broadband mode monitors to the optimized design and resimulate it. Then plot the field distribution for a nice visualization and the broadband transmission as the bandwidth is usually what people care about a lot. 
3. Could you also comment on the total number of simulations ran in the entire optimization process and how long it took roughly speaking? This way users can gauge the efficiency and the cost better. 
4. We also need to add metadata to the notebook files. You can follow the guide [here](https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e) but let me know if anything is unclear. 

And to answer your questions:

1. Yes I think we should. We can keep a few iterations after it plateaus. 

2. Like Tyler suggested, usually we store the hdf5 files in the `data` folder.

3. Don't recall exactly but yeah we should either get rid of it if it's not needed or add it to the current simulation's auto grid. 

5. We try not to give a precise number as the notebook can be modified and our charge formula can change too.  

6. Note that when linking to another notebook, we usually link to the flexcompute domain one, for example https://www.flexcompute.com/tidy3d/examples/notebooks/YJunction/.

tylerflex (commented):
Thanks @m-bone Looks good to me. The converted notebooks seem to have much fewer sharp bits, which is great. And the new bayesian opt demo is great. Here are all of my comments so far.

### Bayesian Optimization

* [pre 27]

`The most effective way to use the Design plugin is to split the workflow into "pre" and "post" functions which can surround a call to the Tidy3D webservers`

It might be helpful to be more explicit and mention “surround a call to the Tidy3D web servers to run a simulation”.

  `The pre-function typically creates a simulation which the post-function then analyzes.` 

Probably better to say that the pre function returns a simulation and the post function analyzes the resulting data after the simulation is run.

* [27] would it make sense to define the input param as just a single list or array? I suppose we just dont have the ability to handle this as a Parameter yet? If not, might be helpful to rename **params to something like **w_values? Perhaps to make it clearer what’s contained in it?

* Could you put docstrings and type annotations in the fn_pre and fn_post? and wherever else appropriate? Would help with readability.


* [pre 28] In the invdes plugin, we talked about adding some kind of flag “minimize” or “maximize” to toggle whether to minimize or maximize the FOM. Maybe we can add that to MethodOptimize at some point.

* [30] 
  * looks like every iteration is a batch of 1 task. It’s a lot of output and also maybe it would be cleaner to run this edge case as a regular Job?  
  * Also, is there a way to get the optimizer state at each iteration out from the design plugin run? Might be nice for there to be a way to monitor progress.
* [43] Results are nice 👍
  * Could we also plot the intensity pattern of the light? With `SimulationData.plot_field()`? Would be nice to see the field pattern.
  * What’s with the small white gap between the polyslab and the waveguide? Is that a plotting artifact only? Is there a way to adjust the vertices x locations slightly to remove this? Might get some questions about it.
  * Would be nice to link at the end (and / or the beginning) to the other optimization notebooks, such as GA and particle swarm if users want to check out more.

### Bullseye PSO

Curious why the field pattern and spectrum changed so much in [13] from the original notebook?
Same comments apply about the type annotations, docstrings, and linking to other notebooks (

### GA

What happened to the 3D plot in [8]?
```
sim = make_sim(pixels)
sim.plot_3d()
```

There’s a deprecation warning that maybe should be addressed after cell [10]

tylerflex:
> Should we using early stop for the GA and PSO? Both have long plateaus at the end. It would demo early stop and it would reduce the notebook runtime

would complicate the notebooks, but maybe we should, considering how long they take to run. I think it's probably ok as is though.

> DesignSpace can easily set the local output directory for sims: should we use one? My "notebooks" directory is full of .hdf5

Yea probably better to set the `path_dir` to `data`.

>  Why is mesh_override defined in cell 7 but never used? Can it be removed?

probably it was meant to be added to the `Simulation.grid_spec.override_structures` but was forgotten? probably should add it.

> Cost warning at the start is "larger than 10 FlexCredits" in every notebook. Should we be more specific or is > 10 fine?

Not sure, probably fine.

> GA notebook needs to link to the BO notebook at the start once the BO notebook is on [www.flexcompute.com/tidy3d/examples/notebooks/](http://www.flexcompute.com/tidy3d/examples/notebooks/)...

tylerflex:
@tomflexcompute maybe we should add a design plugin tag to these examples as well?

tylerflex:
other thing Is we should also update Design.ipynb and ParameterSweep.ipynb to use the new Design plugin (and note any API changes as Notes `> ` in the notebooks).

e-g-melo (approved):
Hi @m-bone ! Thanks for this work! It is very exciting!

I have no additional comments. The notebooks look great to me. 

I would suggest you plot the object function values at each iteration to help users visualize the optimization progress.

I agree that including an early stopping criteria would be very helpful.

m-bone:
> It seems like our `MethodParticleSwarm` is using `pyswarm` under the hood? Even in this case, not sure if we need to expose it explicitly in the notebook?

We could hide it but I've included the relevant packages in the GA and BO notebooks as well. My reasoning is that the package docs have additional description and examples on how different arguments work. The Methods for the optimizers also have many links to the external docs so it's not hidden in the plugin either.

On linking the BO notebook from other notebooks - as the BO notebook is new it doesn't have a link at flexcompute.com/etc. Is the best way to go about it to merge this PR and then make a little PR with the links later once the notebook is live online?

tomflexcompute:
> On linking the BO notebook from other notebooks - as the BO notebook is new it doesn't have a link at flexcompute.com/etc. Is the best way to go about it to merge this PR and then make a little PR with the links later once the notebook is live online?

Yeah this is fine too. The URL is quite predictable however so you can already tell what it is going to be like before it's actually there. Essentially https://www.flexcompute.com/tidy3d/examples/notebooks/`NotebookFileName`/.

m-bone:
I've made all the changes / answered questions raised in the review so far. Additionally, I've updated the other Design plugin notebooks so they're all consistent.

Any changes to the optimization performance are due to differences in the way the random numbers have been generated by each optimization. I've gone through and found seeds that give interesting looking optimization progress and fit the narrative and theme of the notebooks. There shouldn't be a need to redefine how the random numbers are generated (unless the optimizer packages update significantly) so these notebooks shouldn't change significantly between tidy3d versions. 

Feel free to have a look over them again and I'll do a final check once the Design update PR has been merged.

tomflexcompute:
Thanks @m-bone for the updates. Looks very good!

Instead of using the same Y branch schematic image, maybe we can use a new one here to avoid duplication. What do you think about the following image? 
image

Also note that "applications" and "features" tags haven't been added to the Bayesian optimization notebook metadata as described in the last part of this[ instruction](https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e). Sorry for the confusion here since these two fields are newly added so notebooks in your branch don't have them yet, but they do in the develop branch. You only need to add them to the Bayesian notebook I think. 

Everything else looks good to me. Thanks again for the great effort!

m-bone:
Thanks for the image, it's sweet! I've changed it over now.

Apologies, I'd missed that part of the doc; I've updated the metadata now as well. Thanks for the feedback!

Design.ipynb
tylerflex, line 1: in cell [29] there used to be a section using the `batch_data` where `results.batch_data` was called and some processing was done.

It seems this was removed in the new version, so I think we should just include some note saying that this feature was removed and if the user wants to grab the simulation data from the batch, where to find it (preferably also some example showing this).

tomflexcompute:
@m-bone could you also add "Global optimization" to the "features" field in the metadata? Everything else looks ready to go for me.

PR #160, opened 2024-08-20
Small changes to vortex notebook

momchil-flex:
Note: this PR is not necessarily meant to be merged (and it definitely needs to be edited if it is to be merged since it contains two copies of the local projection + plot). But I was playing with the vortex notebook and I noticed a few things:

- Somewhat unnecessarily complicated definitions of distances along z between source and pml, source and structures, monitor and structures. In principle we just need a buffer between the structures and the pml. The source and the monitors can be pretty much anywhere within the buffer region. I'm pretty sure this comes from some of our other examples (maybe in some cases it makes sense e.g. if structure is not centered at 0).
- In the xy size, there is a comment `the size of the metasurface plane as a function of the number of unit cells and the lattice constant a 1.5 lattice constant is added on each side to avoid structures touching the **PML**`, however an extra length of 12 is added in that line without further explanation. This does make the warnings about structures being close to PML go away, but there may be a case to be made that it's better without it.
- The window_size of `0.8` struck me as quite significant since this basically means that almost the entire near monitor is windowed in that direction. It does look like a relatively large size is needed because of the PML effects (although if the extra padding of 12 along x and y is removed, the window size can also be decreased). After playing with that a bit I decided to try using a GaussianBeam instead of a PlaneWave, and that produces a nice looking result even without any windowing - maybe it's a better approach? In fact the GaussianBeam result without windowing looks very similar to the previous windowing result.


FilipeFcp:
@momchil-flex Thank you for the comments. I will take a look at it.

FilipeFcp:
@momchil-flex Can I take this opportunity to ask a question?

The complex definitions of the z-position likely come from my undergrad days using Meep. I recall reading that sources could induce reflections in the simulation, so I typically placed them as far as possible from the PML and structures. Is this a misconception of mine? Is there any issue with positioning the sources near structures or the PML?

Regarding the far-field monitor, is it the same? Can they be placed close to structures, or should they be positioned at least far away to avoid evanescent fields?

For your other comments, I believe the Gaussian beam resolves all issues. With no strong fields near the PML, we don’t need the extra buffer or the `freq_window`. I used a large `freq_window` because, in the experiment, the entire substrate was covered by gold, and only light passing through the metasurface was transmitted. I thought simulating a lot of gold would be computationally expensive. Not sure why I didn’t think of the Gaussian source.

Anyway, I've cleaned up the notebook and added a comment to justify why we can ignore the warnings. Can I push the changes here, or should I open a new pull request?

Thanks!

momchil-flex:
> @momchil-flex Can I take this opportunity to ask a question?

Always! :)

> The complex definitions of the z-position likely come from my undergrad days using Meep. I recall reading that sources could induce reflections in the simulation, so I typically placed them as far as possible from the PML and structures. Is this a misconception of mine? Is there any issue with positioning the sources near structures or the PML?

The only thing I can think of where this would be an issue is a point dipole source. This has evanescent fields which can create gain in the PML if it's too close to the PML, and also it will be influenced by nearby structures that change the LDOS. However, for our directional sources like plane wave, gaussian beam, mode source, it shouldn't matter as long as they are let's say at least two pixels away from PML and structures to avoid some interface pixel effects.

> Regarding the far-field monitor, is it the same? Can they be placed close to structures, or should they be positioned at least far away to avoid evanescent fields?

Again, a couple pixels should be sufficient. The field projection with `far_field_approx=True` does not propagate evanescent fields and those will have no effect. The projection without the far field approximation should correctly propagate the evanescent fields too. In the end, the results should be the same at the same observation points, except potentially some numerical differences that should be quite small.

> For your other comments, I believe the Gaussian beam resolves all issues. With no strong fields near the PML, we don’t need the extra buffer or the `freq_window`. I used a large `freq_window` because, in the experiment, the entire substrate was covered by gold, and only light passing through the metasurface was transmitted. I thought simulating a lot of gold would be computationally expensive. Not sure why I didn’t think of the Gaussian source.

That's fine, these things are not obvious. :) I thought about it because we have had similar discussions before for other problems...

> Anyway, I've cleaned up the notebook and added a comment to justify why we can ignore the warnings. Can I push the changes here, or should I open a new pull request?

Here is fine but for now just push a new commit rather than squashing yours into mine.

FilipeFcp:
Thank you very much for the answers!

momchil-flex:
Looks good, I guess in some sense the question is do we even need to show the windowing function here given that the gaussian beam source does the trick. I guess we could leave it just as an illustration of what can also be done. Maybe to avoid this warning, you could set `colocate=False` in the corresponding near-field monitor and put a comment that it is more accurate to do user-side projections using monitors on the Yee grid?

image

alec-flexcompute (approved):
Hi guys, sorry that it took so long for me to catch myself up with this conversation. It looks like things were adequately addressed and the changes look good - I also think it's instructive to leave the windowing function as is. The warning message could also be instructive, but I can understand wanting to get rid of it for an example. Maybe describe this in the same comment about user-side projections?

FilipeFcp (approved):

FilipeFcp:
Thanks @alec-flexcompute.
I think is all set @momchil-flex.

momchil-flex:
I'm a bit confused, not sure if something is wrong with github or with your last commit, but now I'm seeing very few differences when I go to "Files changed" here. Can you take a look and see if it looks good to you or not?

FilipeFcp:
I checked the file, and it seems correct. I can see the modifications we both made (along with a lot of metadata changes) when I click 'Load diff,' so it seems fine.

PR #163, opened 2024-08-29
Remove the last duplicated metadata

tomflexcompute:
The metadata of the adjoint plugin and autograd notebooks are duplicated. The previous PR has removed most of them but this particular notebook was missed. This PR removed the last bit of metadata duplication.


alec-flexcompute (approved):
looks good, hard to see anything that could go wrong these changes!

PR #164, opened 2024-08-30
Mid-IR Metalens Bug Fixes

m-bone:
Noticed a couple of little bugs in the metalens notebook whilst working alongside it for metalens surrogate notebook. Changes below; they shouldn't be affected by the Tidy3D version so can be implemented when ready.

- Improved performance of metalens by recalculating theta in cell 10. Previously theta ranged between 0 - pi instead of -pi - pi.
- Fixed bug in cell 17 where ylabel was set twice instead of xlabel.
- Fixed typo in metadata "title".


tomflexcompute (approved):
Thanks @m-bone . Good catches!

PR #166, opened 2024-09-04
photonic crystal inverse design notebook (flux / cylinder)

tylerflex:
adds a notebook demo-ing `FieldData.flux` and `Cylinder.radius` differentiation.


tylerflex:
@momchil-flex could you take a look at this and see whether this makes sense as a basic PhC optimization? and whether I should do any modifications?

Thanks

tylerflex:
@momchil-flex just bumping this again so it doesn't fall through the cracks. Wondering if you have thoughts on the specific objective used here and whether I can do something a bit more relevant or if it's fine. Thanks

momchil-flex:
So should I understand that the bend itself (without the straight section) was too good already even unoptimized? Here the starting structure just seems chosen to make the optimization more extreme, but it really is not a great starting structure if your target is to bend the propagation.

I'll think about some meaningful PhC application.

tylerflex:
> So should I understand that the bend itself (without the straight section) was too good already even unoptimized?

Yea without the straight section it was bending fine, but with the straight section, it was a 50% split. I agree we can probably do better. Would there be any possibility to use ModeMonitor and do some dispersion engineering?

tylerflex:
@momchil-flex I reworked this notebook a bit to make it simpler. Goal now is just to optimize transmission through a PhC waveguide, including minimizing reflection at the interfaces

image

image


image


Seems to do pretty well.

Let me know if you think this is a decent enough application and if i should make changes.

momchil-flex:
Ah, so this now does remind me of something that is actually a research topic in photonic crystals: how to couple light in and out of a photonic crystal waveguide from a *regular* waveguide, not from free space. Here's a review I found with a quick search. https://www.academia.edu/download/91638147/j.photonics.2016.04.00120220927-1-1lczr2s.pdf

I think this could be a cool illustration yeah, it's almost like what you've done already. In principle, you don't need to go the full in and out. You can do something like this:
image 

but I guess the problem is that it's hard to put absorbing boundaries over the photonic crystal (even Absorbers seems to cause relfections). Maybe worth a try though? Or alternatively you could put a taper region on one side, then the regular waveguide for some periods, then an output taper on the other side which could just be mirroring the input taper modifications rather than be freely defined?

tylerflex:
Could you send the link again? (or just paper name?) can't open link.

I think i solved the absorber reflection problem, just needed like 4-5 unit cells and 100 or so absorber grid cells but it seems to work pretty well in my bend example

momchil-flex:
Ah sorry: https://scholar.google.com/scholar?cluster=2740535085652940432&hl=en&as_sdt=0,5

tylerflex:
Ok I think this is ready @momchil-flex . but it will only work properly in 2.7.7 with my recent cylinder fix. so i'll only merge assuming we wont do any `latest` updates before then

momchil-flex (approved):
Great! A few small things I noticed:
- `pillars` everywhere should be `holes` now (and note that in one comment it's spelled as `pill|ars`
- This comment also seems stale (no bend anymore)     """Function to generate a simulation, with or without bend, and different monitors depending on whether optimizing."""
- For efficiency: probably no need to have the source being inf along z, and the optimization monitor being inf along y and z, although it may be a small difference.
- Since you just have a regular waveguide, using the default boundary conditions should be fine. Just remove the `boundary_spec` definition: I actually got a much better normalization value (1.000006571892769) because the Absorber was probably introducing a bit of reflection.

PR #167, opened 2024-09-06
uploading ME notebook

FilipeFcp:
Hi all. I am opening the PR for the multipole expansion notebook. Please take your time to review it. 
In cell 7, I used the `eval` function to reduce code redundancy. I'm not sure if this is a good programming technique, though. If you think it’s problematic, I can revise it.


e-g-melo (approved):
Hi Filipe! Congrats, you got an impressive match between analytical and numerical results!

I have only a few comments:
1. I suggest not importing `jupyter_black` because most users should not have it installed.
2. I recommend revising the introduction in order to address the fact that you pointed out the usefulness of multipole expansion in understanding the behavior of particles with arbitrary geometries. However, the example is based on a sphere. Maybe you can explain that you will use a simple sphere here to be able to validate the results using Mie theory.
3. I think most of the users will try the example on a computer with less than 32 GB RAM. Besides, our notebook environment supports only 4GB. Would it be possible to alleviate the memory usage? Maybe by reducing the wavelength range and using fewer frequency points.
4. I suggest to use `freq0 = (freq1 + freq2) / 2` in replacement of `freq0 = (freq1 - freq2) / 2 + freq2`
5. It is not clear which medium you are refering to here: "where n is the medium index, which is 1 in this example"
6. I suggest explicitly defining all the parameters in the intensity definition before cell [3].
7. Maybe you can move the geometry definition to cell [3], close or within the structure definition.
8. "the function to calculate the the multipole expansion" -> "the function to calculate the multipole expansion".
9. I've used the `eval` function before, and Momchil recommended me to avoid it. But I am not sure if it is critical here.

FilipeFcp:
Thanks @e-g-melo.

I've addressed almost everything.
Regarding the geometry definition (number 7), I'm inclined to keep it in the first cell, as it makes it easy to change for any arbitrary geometry.

tomflexcompute (commented):
Thanks @FilipeFcp . Very nice demo that will be referenced by many users in the future. 

1. Agree with Emerson that we should remove jupyter_black in the notebook since most users probably don't have it installed. As long as we format the code before merging, it should be fine.  

2. I suggest making a schematic image for the thumbnail. For example, we can make a sphere, use some wave to indicate the incident and scattered field, and then maybe show the multipole fields as inset. 

3. A typo:
"pos-processing" -> "post-processing"

alec-flexcompute (approved):
Wow, this is a very impressive notebook - the agreement with the analytical results is great! A couple (small) things:
1. In the intro, do we need the discussion of what to do about divergence?
2. Before cell 3, when it says "This source injects exactly 1W/um^2..." maybe take the 1 out of the numerator
3. It might make things clearer to give the ME function in markdown before it's given in code (before cell 7)
4. Above cell 12, Plotting with two ts :)

FilipeFcp:
Hi all, I believe it is good to go now.

I made some changes based on Monchil's comments regarding the source and monitor positions. I also added the data files from PieMieScatt instead of calculating in the code, since the library uses an outdated version of SciPy.

Thank you @alec-flexcompute and @tomflexcompute , sorry to take so long to reply to you, but I have addressed all your comments.

tomflexcompute:
Thanks @FilipeFcp . Looks very good. A few more minor comments before merging for you to consider:

1. I feel like the title can be more elaborate. Maybe something like "Multipole expansion for electromagnetic scattering".
2. The image size is a bit large. Can consider setting the size to a smaller value like 400.
3. It looks like you haven't installed `ipywidgets` and thus the warning in cell[5]. Can you install it (`pip install ipywidgets`) and rerun the notebook?
4. In the last cell you printed the tidy3d version, which I think can be deleted.

FilipeFcp:
Hi @tomflexcompute. Thanks for revising it again.

PR #172, opened 2024-09-20
correct links

FilipeFcp:
I found a broken link in the Nanobeam notebook and got curious, so I ran a sweep through all the notebooks and found a few more. I’ve corrected them.


alec-flexcompute (approved):
Looks good, thanks for the check!

tomflexcompute (approved):
Thanks Filipe.

PR #176, opened 2024-10-04
:memo: Basic Example section introductions

daquinteroflex:
This will not be fully implemented as intended, but it's a small interim step that at least provides some availability of the content whilst the extension functionality eventually works.

https://github.com/flexcompute/autoflex/issues/10


daquinteroflex:
Wrong branch naming

PR #184, opened 2024-10-24
Added the metalens waveguide taper example

tomflexcompute:
This notebook demonstrates the design of a metalens-assisted waveguide taper and then optimized with inverse design. Currently if using `symmetry=(0, -1, 1)` will cause some issues (due to half of the structures being outside?) so I only use (0, 0, 1).


MetalensWaveguideTaper.ipynb
yaugenst-flex, line 1: This is super cool!

Minor comments:
 - "we will explore the concept of designing **an** on-chip metalens"
 - "Our goal is to design a metal-assisted waveguide taper" --> probably metalens-assisted?
 - "tuned accordingly to the specific fabrication constraint" --> according to

One potential issue I see with the optimization is that the `tanh` projection here isn't really necessary and might actually bias the optimization because it introduces a nonlinearity in the length of the elements. Because of the `tanh`, lengths in the middle of the range (0.5) have a much stronger gradient than ones at the outer edges of that range, which introduces a bias into the optimization away from "medium-sized" elements.
Instead, it might be a better strategy to optimize the lengths directly and then constrain the parameters to stay within the allowed range, i.e. in the optimization loop:
```python
params = optax.apply_updates(params, updates)
params = np.array(params)
params = np.clip(params, l_slot_min, l_slot_max)
```

The learning rate will then also directly reflect the maximum physical length that a metalens element is allowed to change during each iteration, which might be more intuitive too.

yaugenst-flex:
> Currently if using symmetry=(0, -1, 1) will cause some issues (due to half of the structures being outside?) so I only use (0, 0, 1).

@tomflexcompute this is fixed in flexcompute/tidy3d#2025

tomflexcompute:
Thanks @yaugenst-flex for the great feedback. All are incorporated in the current version. 

Optimizing the lengths directly does work. The end result is somewhat similar to the previous approach but I do agree that this is a more intuitive and convenient way.

yaugenst-flex (approved):
Looks great!

tylerflex:
Thanks @tomflexcompute  this is great! just a few comments:

1. Do we want to call this `Autograd20....ipynb` just to keep all of these in the same place? or is it becoming too many now?
2. The symmetry can be handled by just cutting the number of design parameters in half and simply setting the + and - y structure. I think it should work automatically, and after Yannik's last PR to develop. Might be worth a try since it can be a more tricky aspect of doing optimization so good to have an example of it.
3. There is also a `Structure.background_permittivity` to set the background perrmitivity for shape optimization. This is really only needed for PolySlab and Cylinder at the moment because Box automatically grabs the permittivity, but it might be good to add it `background_permittivity=si` in case users want to change the geometry. it's often the safer approach.

Both 2 and 3 are just enhancements though, overall it looks great!

tylerflex (approved):

tylerflex:
Oh yea and dont forget to add this notebook to the RST file so it shows up in the docs

tylerflex:
Also forgot one comment: similar to the grating coupler one you recently did: what happens if you just start with a uniform design and let it run? In principle no need to look up the optimal designs from the paper, just would be curious

tomflexcompute:
Thanks @tylerflex for the great feedback. Where is this `background_permittivity`? I can't find anything in the docs.

tylerflex:
Background permittivity was just merged into develop. So yea actually we can wait to include it after 2.7.6. I can handle it

tomflexcompute:
Great @tylerflex . All comments addressed besides the `background_permittivity`.

Also tested the optimization starting from a uniform slot length. After 30 iterations the FOM is at about 0.5. Maybe eventually get to a good place if running for, say, 100 iterations but that would take too much time. Starting from a reasonable design is still much better.

e-g-melo (approved):
Hi @tomflexcompute! 

Thanks! The notebook is very didactic and well structured. It is a very nice and simple example on shape optimization. I have no additional comments.

FilipeFcp (approved):
Hi @tomflexcompute. This is really cool. I like these applications that leverage the adjoint method combined with physical insights.
I also don't have any additional comment.

PR #185, opened 2024-10-26
Autograd quantum light extractor notebook

tylerflex:
see https://github.com/flexcompute/tidy3d/issues/2042


tylerflex:
@e-g-melo I cleaned up this notebook and it should be ready to merge. Would you mind taking a look through beforehand? Thanks

e-g-melo (approved):
Hi @tylerflex! Thanks for these improvements.

1. Regarding your comment: # Shouldn't this be --> |x-x0|^2 + |y-y0|^2 <= r*2
Yes, that is true. I thought it was already corrected, but somehow I missed this. 
I think we can increase `non_etched_r` to 0.06 to avoid changing the results.

2. We need `plt.show()` in cells 20 and 21. 
3. I wonder if we could remove cell 30.

PR #186, opened 2024-10-28
add lumped element tutorial

dmarek-flex:
Add tutorial for the new linear lumped elements for RF and microwave applications. Still rough draft, since waiting on branch to be merged into pre/2.8 before final run of the notebook. 

- [x] replace local running with web api run
- [x] need new  feature image
- [x] rerun without temporary deployed solver


tomflexcompute (approved):
Very nice @dmarek-flex . All look very good. Just one question: you mentioned the discrepancy is due to the current calculation not including the effect of the displacement current. If you use a finer grid, the discrepancy is the still same?

I haven't checked yet but can you add metadata to the notebook if not already?

dmarek-flex:
> Very nice @dmarek-flex . All look very good. Just one question: you mentioned the discrepancy is due to the current calculation not including the effect of the displacement current. If you use a finer grid, the discrepancy is the still same?

Good question, I had not thought about it, but I just tested and I will add a comment on the notebook. The discrepancy is reduced, because the path integral becomes smaller, reducing the enclosed displacement current, while still enclosing the same amount of conduction current. 

> I haven't checked yet but can you add metadata to the notebook if not already?
Should be added.

dmarek-flex:
I doubled the resolution and get this plot

image

weiliangjin2021:
> The discrepancy is reduced, because the path integral becomes smaller, reducing the enclosed displacement current, while still enclosing the same amount of conduction current.

What if you just increase the resolution, but apply the same path integral?

weiliangjin2021:
Looks great. A few comments:

- Split cell 1 into two cells: 1st cell for importing all modules, and setting log levels; 2nd cell for setting up parameters.
- Cell 3: "#units are Henrys"  --> Henries (wikipedia says "The United States National Institute of Standards and Technology recommends users writing in English to use the plural as henries")
- The 1st plot below Cell 4: would be good to use plt to add text to the figure explaining which end is port and which the lumped element.
- Below Cell 6, a brief discussion on why the field looks correct?
- Below Cell 7: a brief and intuitive description of why the S11 plot looks expected.
- Regarding the deviation at high frequency, it might also be related to radiation to the farfield from lumped element, which is equivalent to add additional lumped elements. But maybe it's equivalent as your displacement current explanation, as the radiation can originates from the displacement current.
- In the 2nd example, I remember you mentioned that S parameter is different from Tafloev' book. Is that still the case?

dmarek-flex:
Thanks @weiliangjin2021 I modified the notebook accordingly, see my responses to some questions:

> * Below Cell 7: a brief and intuitive description of why the S11 plot looks expected.

Not easy to give intuitive description, so I added section computing S11 from transmission line theory!

> * Regarding the deviation at high frequency, it might also be related to radiation to the farfield from lumped element, which is equivalent to add additional lumped elements. But maybe it's equivalent as your displacement current explanation, as the radiation can originates from the displacement current.

 I think it is due to the displacement current calculation, I still need to try the idea you had comparing different discretization with the same path integral definition.

> * In the 2nd example, I remember you mentioned that S parameter is different from Tafloev' book. Is that still the case?

The scattering parameters do not match Taflove's book, but I think there are details of the simulation setup in Tfalove's book and in the original reference that are missing or incorrect. I am very confident in the computed S parameters that I am getting. Especially, now with the additional validation using transmission line theory.

weiliangjin2021 (approved):

QimingFlex:
@dmarek-flex Looks great!  I have a few questions about the benchmark:

1. To ideally match the transmission line structure, should the side walls ideally be replaced by PMCs? The field distribution shows an obvious fringing effect, which might impact accuracy.  This might not be important as the convergent result agree well with the analytical solution.

3.I tried using a uniform mesh with `grid_spec=td.GridSpec.uniform(dl=25)` and got an error, which might be a separate issue. 
Screenshot from 2024-11-04 20-37-26

4. Why was 4 mil chosen as the metal thickness? I checked the literature but didn’t see any mention of thickness; would using PEC_2D be better?

dmarek-flex:
> 1. To ideally match the transmission line structure, should the side walls ideally be replaced by PMCs? The field distribution shows an obvious fringing effect, which might impact accuracy.  This might not be important as the convergent result agree well with the analytical solution.

That might make an interesting test case, but this example is a more realistic transmission line, which might be easier for users to understand.

> 3.I tried using a uniform mesh with `grid_spec=td.GridSpec.uniform(dl=25)` and got an error, which might be a separate issue. Screenshot from 2024-11-04 20-37-26

Forgot to consider grid snapping and how it effects the placement of the lumped element monitor. I'll investigate.

> 4. Why was 4 mil chosen as the metal thickness? I checked the literature but didn’t see any mention of thickness; would using PEC_2D be better?

Yea, I think I just changed that part of the structure. I am not trying to completely reproduce the structure in Taflove. I think I chose a finite thickness to simplify the overall problem.

tylerflex:
just checking in on the open PRs, is this one good to merge?

dmarek-flex:
> just checking in on the open PRs, is this one good to merge?

Still trying to finalize things with lumped elements, which might change this tutorial

dmarek-flex:
> 3.I tried using a uniform mesh with `grid_spec=td.GridSpec.uniform(dl=25)` and got an error, which might be a separate issue.

I should have tried this a long time ago, but I am currently unable to reproduce this problem using pre/2.8. Do you still have this problem @QimingFlex ?

dmarek-flex:
> I should have tried this a long time ago, but I am currently unable to reproduce this problem using pre/2.8. Do you still have this problem @QimingFlex ?

I was able to reproduce this issue after trying different values for `dl`. This problem is rooted in the snapping behavior, so it is best not to use the UniformGrid type with lumped elements. Using AutoGrid should resolve these types of issues.

Because of that problem, and others, @weiliangjin2021 is working on improved meshing which should hopefully remove snapping in most cases, since it should really be avoided in the first case.

QimingFlex (approved):

PR #189, opened 2024-10-30
Kerr sideband example notebook

FilipeFcp:
Hi all,

This example was suggested by the user who was experiencing convergence issues with chi3 simulations.
He did most of the work; I reviewed it, added the image, and included the section on iterating through amplitude parameters to analyze convergence.

It might be a good idea to test it with the new @caseyflex method before publishing.


alec-flexcompute (commented):
Hello Filipe,

This is a very informative example with great results - thanks for doing it! Just a couple of thoughts:

- Maybe at the beginning to a quick intro of what sidebands are so the user can use this notebook as an explanation for phenomena they're seeing
- In the total field amplitude expression, maybe specify why the exponentials have opposite signs. Is that so the field power can be expressed in terms of w_d?
- I think you can probably suppress errors and make a note that you're doing that (especially for cell 45)

Other than that, it looks great to me!

caseyflex (changes requested):
This is a really great demo notebook!

I actually disagree with @alec-flexcompute 's last point -- we probably should not encourage users to suppress errors or warnings in general, since we can use those to communicate important information. Instead, we should change the setup to remove the warnings, or explain why the warnings that do appear can be safely ignored. In this case, perhaps we can make simulation `shutoff=0` to suppress the warnings about field decay. 

I also see solver warnings about the source time dependence for broadband source, which are a bit more concerning. @dbochkov-flexcompute is this source decay an important requirement for broadband sources? And actually now that I think about it, CW excitation is anything but broadband -- you should probably set `num_freqs=1` in the mode sources for a much more efficient simulation.

I am in the process of deploying and testing to see if we can improve the convergence. I will let you know how that goes.

I see that the fit in cell [46] gives an exponent of 4.8, which is about 5, which I think is reasonable from the other equations. Maybe we should just fit to a*X**5, since fitting the exponent can be a bit sensitive? Although I am also fine with it as is, but maybe add some discussion about the exponent.

Thanks!

caseyflex:
Oh, also, maybe you can use a simulation Batch for the amplitude sweep? and add a note about the cost of running the whole notebook.

Also, the simulation name "ring resonator" should be changed, and there is some old commented out code that can be removed. The name "kerr_sidebands_%i"%amp" appears to not work since amp=400. The plot label in cell [46] is a bit messed up. And I'd change "exponential form" -> "power law".

caseyflex:
Could you also add a note about why `NonlinearSusceptibility` is used here instead of `KerrNonlinearity`? I realize the difference is technical and our naming is not the best -- it would be good to emphasize that the real-fields nonlinearity is the right one to use here.

caseyflex:
I ran a few more data points with the new nonlinear improvements. It no longer diverges; I am still not sure if the results are physical for strong nonlinearity. Obviously it no longer has the power-law form, but I see this is also derived under the assumption M << sqrt(2). If the results are not physical, this can sometimes be improved by decreasing the Courant factor, but a quite significant decrease could be required.

image

FilipeFcp:
Hi @alec-flexcompute and @caseyflex. Thank you for the review. I removed the shutoff, so most of the warnings are gone.

I’m still not sure about the difference between `NonlinearSusceptibility` and `KerrNonlinearity`. Does one implement real fields while the other uses complex fields? Just out of curiosity, I ran the simulation again with `KerrNonlinearity`, using the n2 value, and the results are pretty much the same. The only thing is that the simulation for amplitudes around 700 still doesn't diverge.

image

The results with the Newton method seem nice. It’s good that it won't diverge anymore. However, whether the results are physical is a more delicate matter. This chi2, chi3 approach, if I remember correctly, is only perturbative, so even in real life, it would be valid only for relatively small intensity regimes. I will add a brief discussion about that.

Should we wait for the new solver version the be merged in prod to publish the example? Or go as is, and them change in the future?

caseyflex:
> Hi @alec-flexcompute and @caseyflex. Thank you for the review. I removed the shutoff, so most of the warnings are gone.
> 
> I’m still not sure about the difference between `NonlinearSusceptibility` and `KerrNonlinearity`. Does one implement real fields while the other uses complex fields? Just out of curiosity, I ran the simulation again with `KerrNonlinearity`, using the n2 value, and the results are pretty much the same. The only thing is that the simulation for amplitudes around 700 still doesn't diverge.
> 
> image
> 
> The results with the Newton method seem nice. It’s good that it won't diverge anymore. However, whether the results are physical is a more delicate matter. This chi2, chi3 approach, if I remember correctly, is only perturbative, so even in real life, it would be valid only for relatively small intensity regimes. I will add a brief discussion about that.
> 
> Should we wait for the new solver version the be merged in prod to publish the example? Or go as is, and them change in the future?

I see, that's good then. NonlinearSusceptibility has a nonlinear polarization proportional to `E^3` and uses E real (or applying separately to real and imaginary components). KerrNonlinearity has a nonlinear polarization proportional to `(E * conj(E)) * E` where E is complex. Usually NonlinearSusceptibility would be preferred as it is closer to the underlying physics. But it makes sense that both would capture the effect here as expressed in the NLSE.

Hopefully once you make num_freqs=1, all the warnings will go away...

Up to you whether to wait. It's probably fine to go as is, if people want this example live sooner.

FilipeFcp:
Alright. I pushed a new version addressing these comments.
Thank you again @alec-flexcompute  and @caseyflex

dbochkov-flexcompute:
> I also see solver warnings about the source time dependence for broadband source, which are a bit more concerning. @dbochkov-flexcompute is this source decay an important requirement for broadband sources? And actually now that I think about it, CW excitation is anything but broadband -- you should probably set `num_freqs=1` in the mode sources for a much more efficient simulation.

Yes, that's right, we should add a validator to prohibit using broadband with a CW source

tomflexcompute (commented):
Thanks @FilipeFcp for polishing this. Looks very good.

1. Add the notebook title to the .rst file in docs folder. 
2. In the intro, mention the affiliation (and title if possible) of the author to make it more specific such as Dr. Chenchen Wang, postdoc researcher from the University of Wisconsin–Madison.
3. Since ContinuousWave is rarely used (actually I think it's never used in our public examples) and not documented in the docs, maybe adding a few more comments about it would be helpful. Usually we should warn the users to not use it in general in linear sims. Also explain what fwidth here means? Can you use GaussianPulse with a very narrow fwidth in this case?
4. I think the mode profile plottings are rotated by 90 degree. In the `plot` method, I believe you can do something like `plot(x="y", y="z")` to change the orientation.

caseyflex (approved):

alec-flexcompute (approved):

FilipeFcp:
I noticed that some y-labels were not exactly correct since normalization has no meaning for a CW source, so I just added (a.u.) in the y-axis. Also, I added a note about this and the meaning of the `amplitude` argument:

image

tomflexcompute (approved):
Fantastic @FilipeFcp ! All look great. Cell[6] seems to have not been formatted by `black`?

FilipeFcp:
@tomflexcompute  Maybe I changed something after removing the black formatting. I ran it again to make sure.

tomflexcompute:
> @tomflexcompute Maybe I changed something after removing the black formatting. I ran it again to make sure.

Thanks. All look good to go.

PR #190, opened 2024-11-06
Added the new inverse design notebook

tomflexcompute:
- Added the new inverse design notebook of the GaP light extractor for diamond NV centers.
- Removed duplicate on the AdjointPlugin13 notebook.
- Added missing metadata on the Autograd12 notebook.

The current version uses the manual soft min with a relatively large tau due to limitations @yaugenst-flex is aware of and is going to remove later. After the fixes are in we can update this notebook but we can publish the current version for now.


tylerflex (approved):
Thanks @tomflexcompute , this new example is great!

A few comments

[5] I think a note somewhere that you still need to use a `FieldMonitor` and call `flux` instead of `FluxMonitor` would help to cut down on confusion.
[7] define `run_time` at the top level
[11] note that yannick recently introduced a `softmin` function in this [[PR](https://github.com/flexcompute/tidy3d/pull/2055)](https://github.com/flexcompute/tidy3d/pull/2055). So we can use it after the next release or try it here if using develop.
[11] consider putting `lambda dim: np.sum(batch_results[dim]["field"].flux.data)` in a function also maybe `lambda flux: np.exp(-flux / tau)`
[optimization] I think it might be better to reduce the learning rate a bit (maybe 0.1?). Also it could be nice if we could symmetrize the parameters? perhaps by applying `params = np.flipud(np.fliplr(params)) / 4.0` in the optimization?
[final remarks] unfortunately we dont support broadband `FieldMonitor.flux` yet, so will need to change mentions of this.

otherwise i pre-approve

tomflexcompute:
Thanks @tylerflex all are great comments. 

Currently if I reduce the learning rate to 0.1, it will run into the no adjoint source error after a few iterations, probably due to the difference between fluxes happens to be large. I can further increase `tau` but that soft min is already pretty soft. Yannick is aware of this issue so will fix it later. Then we can update it to try a lower learning rate as well as change the soft min to the built-in function. 

Regarding symmetry, this design technically can be symmetric with respect to the diagonal. In principle I can also rotate the source to the diagonal direction by using two perpendicular dipoles with sqrt(2)/2 amplitudes to make the design symmetric w.r.t. x or y but that might seem a bit confusing to users so I opted to use the same approach as introduced in the paper for better consistency. This leads to slight asymmetry even w.r.t. the diagonal which is consistent with the paper result. 

One strange thing I did notice is that the final result is noticeably different sometimes even if I just rerun the notebook without making changes to the parameters. Do you see why that might be the case?

AdjointPlugin13Metasurface.ipynb
```diff
@@ -1719,7 +1719,7 @@
   }
  ],
  "metadata": {
-  "description": "This notebook demonstrates how to perform inverse design optimization of a diffractive metasurface in Tidy3D FDTD.",
+  "description": "This notebook demonstrates how to perform inverse design optimization of a diffractive metasurface in Tidy3D FDTD using the adjoint plugin..",
```
yaugenst-flex, line 1722: I know it's only metadata but I think a superfluous period slipped in at the end.

Autograd21GaPLightExtractor.ipynb
yaugenst-flex, line 1: Looks great! I think Tyler's review pretty much covers everything. I also support imposing symmetry. An alternative to Tyler's approach would be:
```python
x = np.concatenate((x, x[::-1]), axis=0)
x = np.concatenate((x, x[:, ::-1], axis=1)
```
This has the added benefit of reducing the number of design variables by a factor of 4.

> One strange thing I did notice is that the final result is noticeably different sometimes even if I just rerun the notebook without making changes to the parameters. Do you see why that might be the case?

I'm only guessing here, but I suspect this is caused by the same reason there is asymmetry appearing in the first place - possibly a slight error in the discretization, as well as in the gradients, leading to potentially random round-off errors.

I'm working on flexcompute/tidy3d#2061. I guess the question is: do we want to wait for that or merge now? I'm fine with either, but that PR shouldn't take too long.

yaugenst-flex (approved):

e-g-melo (approved):
Hi @tomflexcompute! Great example! Thanks

I have only two comments:
1. I wonder how this `sofmin` function is different from the one used in [this notebook](https://www.flexcompute.com/tidy3d/examples/notebooks/InverseDesign/). Do you know which one performs better?  
2. You have reproduced the paper FOM, but since the dipole power can vary along the optimization, would it make sense to optimize the upward efficiency (`upward_flux/total_flux`) instead of the upward flux?

yaugenst-flex:
> I wonder how this sofmin function is different from the one used in [this notebook](https://www.flexcompute.com/tidy3d/examples/notebooks/InverseDesign/). Do you know which one performs better?

The two approaches are a bit different, where the one from the notebook is a weighted sum using the softmin distribution, and the one introduced by my PR is a smooth approximation of the minimum function. In particular, the notebook approach is a minimum-weighted average, which will never converge to the actual minimum, whereas the smooth minimum approximation _will_ converge to the true minimum for small `tau`, i.e. it is a direct approximation of the minimum function.

Which of these performs better in practice I think is hard to say. The smooth minimum approach is a bit more flexible because you can tune how well the minimum is approximated. Also I believe the minimum approximation is a bit more numerically stable because of the log sum exp trick.

Autograd21GaPLightExtractor.ipynb
tomflexcompute: > Looks great! I think Tyler's review pretty much covers everything. I also support imposing symmetry. An alternative to Tyler's approach would be:
> 
> ```python
> x = np.concatenate((x, x[::-1]), axis=0)
> x = np.concatenate((x, x[:, ::-1], axis=1)
> ```
> 
> This has the added benefit of reducing the number of design variables by a factor of 4.
> 

This is still enforcing symmetry in the x and y directions I think? In this case the symmetry axis is actually diagonal.

yaugenst-flex: Oh sorry I missed that. Maybe `x = (x + x.T) / 2`?

tomflexcompute: Does this actually reduce the number of design parameters?

yaugenst-flex: Nope that version doesn't reduce them. One can probably also find a smart way to tile it and reduce the no of parameters..

tomflexcompute: Yeah exactly that's probably why in the paper they mentioned they avoided using symmetry. Let me try a few things to see if I can make it work but it's also nice to follow the paper as closely as possible so readers can reference the paper and more easily understand the content. Essentially the paper serves as an additional manual for the notebook.

AdjointPlugin13Metasurface.ipynb
tomflexcompute: good catch!

tomflexcompute:
> 2. You have reproduced the paper FOM, but since the dipole power can vary along the optimization, would it make sense to optimize the upward efficiency (`upward_flux/total_flux`) instead of the upward flux?

Thanks @e-g-melo that's a good point. In the final remark I kind of mention this as well. In this case since the dipole is relatively far from the extractor, the purcell effect is very weak, only 10-20%. Also this should be a real effect so it would be good to enhance the dipole power too. The overall goal is just to get the net flux as high as possible.

Autograd21GaPLightExtractor.ipynb
tomflexcompute: > I'm working on https://github.com/flexcompute/tidy3d/pull/2061. I guess the question is: do we want to wait for that 
> or merge now? I'm fine with either, but that PR shouldn't take too long.

That's great. We can merge it as it's now and update it after the update is officially released. It's targeting 2.7.7 or 2.8.0?

yaugenst-flex: The autograd thing will probably make it into 2.7.7, not guaranteed though. Yeah let's merge and then update later. We'll have to make a pass over a bunch of notebooks anyways.

PR #191, opened 2024-11-08
:wrench: Verified correct bullet point structure

daquinteroflex:
Turns out we have to add spaces between the bullet points like [Primer.ipynb](https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/Primer.html). I've tried grepping but hard to match so can catch them as it comes up in the docs.


dmarek-flex (approved):
Thanks!

PR #192, opened 2024-11-08
Tyler/autograd /polyslab fix3

tylerflex:
shape optimization results after polyslab fix 3


yaugenst-flex (approved):
Awesome, both notebooks show a clear improvement (at the same iteration no.)!

PR #197, opened 2024-12-04
Updates to radiative loss notebook

momchil-flex:
I made some small updates to the notebook:
- Running fewer radius points to speed up a bit
- There was an instance of something like `loss.append(loss)` which was just creating an infinitely self-nested dictionary
- In the plane size test, because you were running with a given size, then incrementing by 2, but then storing the size at which the loop exited, the stored size was wrong by 2 (i.e. the loss became positive at a given size x but then x + 2 was stored).
- Fixed some more instancess of capital first letter `Loss`, `Keff`


FilipeFcp (approved):
Great, @momchil-flex , thank you!
I’ll work on improving my Python standards. The loss.append(loss) issue happened because I initially used the list variable as Loss, then changed it to loss but forgot to update the other variable...

Anyway, thanks again for your review!

momchil-flex:
Sounds good! Before I overwrite your notebook, do you want to save that version in case you want to e.g. present the figures with more points on them in the meeting next week?

FilipeFcp:
I do have a copy already. Thank you!

PR #200, opened 2024-12-11
Fixed angle

dbochkov-flexcompute:
Demonstration of fixed angle feature. Any volunteers to try it out on a more practical case? @tomflexcompute @e-g-melo @FilipeFcp


tomflexcompute (changes requested):
Thanks @dbochkov-flexcompute for this highly anticipated feature! This tutorial is great as well. We will definitely make some more examples using it, such as metasurface designed for oblique incidence.

A few comments:

1. For the title, I wonder if something like "Broadband plane wave with a constant oblique incident angle" would be more intuitive. 

2. In the intro, you mentioned "Tidy3D provides two methods for simulating angled plane waves". I think technically we provide 3 methods since there is TFSF as well. So to be more specific here we can say something like "two methods for simulating angled plane waves for periodic structures".

3. In the `Simulation Cost` section it's kind of discussed already but to make sure readers understand the context more explicitly, it might be worth mentioning that using the plane wave with bloch boundary to simulate a broadband response at constant angle, one can create a parameter sweep of the source central frequency. Essentially treating the simulation as a frequency domain simulation. Since tasks are run in parallel, the speed might not actually be much worse but the cost linearly scales with the number of sampled frequency points. 

4. A minor cosmetic thing: in the last plot, we can use dark blue solid curve for Tidy3D fixed angle and light blue dashed curve for TMM theta=theta0. Similarly use dark red and light red for example for the bloch boundary ones. This way it might be visually easier to tell which curve is supposed to be compared to which curve.   

5. Could you add the notebook to `tidy3d-notebooks/docs/features/sources.rst`

Some minor typos:

- "increase depends depends on" -> "increase depends on"
- "boudnaries" -> "boundaries"
- "opposites" -> "opposite"
- "continious" -> "continuous"
- "of the injected plane wave. Which means that only" -> "of the injected plane wave, which means that only"
- "dependece" -> "dependence"
- "computaional" -> "computational"
- "dieletric" -> "dielectric"
- "explicity" -> "explicitly"
- "quantative" -> "quantitively"

dbochkov-flexcompute:
@tomflexcompute thanks for the excellent suggestions and corrections! adopted them all

momchil-flex (approved):
Great stuff and good comments by Tom already. Looks good to me, just one very pedantic clarification

> In the intro, you mentioned "Tidy3D provides two methods for simulating angled plane waves". I think technically we provide 3 methods since there is TFSF as well. So to be more specific here we can say something like "two methods for simulating angled plane waves for periodic structures".

Technically, TFSF is also compatible with Bloch boundaries if extruded through them. There's two specific use cases: one is when the simulation is periodic along one dimension but not the other. The other is when the simulation is fully periodic but you still want to get the TFSF effect above the TFSF box (if e.g. injecting from below), i.e. have zero field in the scattered field region if there is no scattering.

Actually now that I think about this, the distinction is not really Bloch/fixed angle/TFSF, it's more Bloch/fixed angle and PlaneWave/TFSF. Only PlaneWave supports fixed angle currently, but in principle it could be some day added as an option to TFSF. In fact I think the Taflove definition of TFSF is a fixed angle one, but we went with essentially a Bloch definition (the auxiliary simulation is always Bloch, even if the main one has PML) because it is much easier to do for a multi-layered backgroudn.

So finally, maybe what we should say is that when using `PlaneWave`, tidy3d provides two methods for simulating angled waves, and mention that the fixed angle is not yet available in TFSF?

dbochkov-flexcompute:
> So finally, maybe what we should say is that when using `PlaneWave`, tidy3d provides two methods for simulating angled waves, and mention that the fixed angle is not yet available in TFSF?

changed Tidy3D Python API section to:
> Oblique incident plane waves can be injected in Tidy3D simulations using source types TFSF and PlaneWave. Currently, TFSF supports only the fixed in-plane k approach, while PlaneWave supports both. Switching between the two techniques for PlaneWave sources is simply accomplished by changing the field angular_spec from FixedInPlaneKSpec (default) to FixedAngleSpec. Additionally, for the fixed in plane k simulations Bloch boundary conditions must be correctly set up, while for the fixed angle simulations periodic boundary conditions are simply set to Periodic.

tomflexcompute (approved):
Fantastic! Thanks @dbochkov-flexcompute

PR #212, opened 2024-12-22
Added the metasurface qBIC example

tomflexcompute:
This PR adds a metasurface example to demonstrate the use of a constant angle broadband plane wave source. 

@e-g-melo we can add this example in the GUI once GUI supports constant angle option for the plane wave. This should just be an additional option in the plane wave settings so should require minimal design?


dbochkov-flexcompute (approved):
Thanks @tomflexcompute, this looks great! Just a few comments:
- a typo in `dx = 0.226  # marjor axis diameter of the nanodisk`
- are we fitting aSi here because our library option doesn't go all the way up to 900nm? maybe add a brief note on that?
- out of curiosity I re-ran the notebook with using pml (I set num_layers=20 randomly), and it seems to be stable, just decays slow
- out of further curiosity, I re-ran the notebook with nk values digitized from the paper figures, and it indeed produces much sharper features. Here the nk file [paper_si.txt](https://github.com/user-attachments/files/18220739/paper_si.txt) in case you want to use it (I used a fit with 2 poles)
Screenshot_20241221_215840

e-g-melo (approved):
Thanks @tomflexcompute! It is a very nice example.

When you plot the simulation grid, maybe you could include information about the resulting grid size compared to the simulation wavelength so that users can understand what would be a sufficiently fine mesh.

I will reproduce it in GUI as soon as the feature is available.

tomflexcompute:
Thanks @dbochkov-flexcompute for the prompt review (though working on the weekend is not encounraged)! All points addressed. 

Yeah using pml will not diverge but the convergence is quite slow and it might actually end up using more credits than absorber with more layers. 

The discretized index is very nice. I was a bit too lazy to try that. I have used it in this version. One thing that I'm not sure if it's of concern is that the qBIC mode wavelength is 900 nm at 16 degree in our simulations. However in the paper it's about 14 degree. Possibly that's just due to the substrate index being slightly different as that's not disclosed in the paper. `n_pyrex = 1.47` seems to be quite generic.

tomflexcompute:
> Thanks @tomflexcompute! It is a very nice example.
> 
> When you plot the simulation grid, maybe you could include information about the resulting grid size compared to the simulation wavelength so that users can understand what would be a sufficiently fine mesh.
> 
> I will reproduce it in GUI as soon as the feature is available.

Thanks @e-g-melo !

PR #216, opened 2025-01-02
Tom/absorber reflection

tomflexcompute:
This PR adds a notebook on benchmarking the reflection from the absorber as a function of `num_layers` since users are sometimes confused about how many layers are needed. 

Also fixed two typos Weiliang found in two inverse design notebooks.


FilipeFcp (approved):
Thanks @tomflexcompute. 
Very nice and informative example. I have nothing to add.

e-g-melo (approved):
Hi @tomflexcompute, thanks for the nice tutorial!

I would only suggest including the comparison between PML and Absorbing boundaries in the title somehow.

momchil-flex (commented):
Great. Two comments about clarifying things further:

- Something related just popped up in an example that @alec-flexcompute was debugging. The reflection of the absorber is likely to increase with increasing resolution, that is to say, more layers are needed for the same physical simulation depending on the grid size. This is because, in some sense, we shouldn't be defining absorber number of layers at all, but actually physical thickness of the asborber. This boundary, unlike PML, is very similar to a physical absorbing region, so if the number of layers is kept the same while the resolution increases, the physical thickness of that region decreases, making the transition less adiabatic, and introducing more reflections. Does that make sense? This also generally suggests that we might want to make an addition to the frontend to be able to define the physical thinkness of the absorber region, rather than the number of layers, such that results do not change unexpectedly with increasing resolution.
- In the final paragraph, where you discuss the possibility that the absorption is not strong enough, you could add a note that this would manifest itself as the reflection level not decreasing when more layers are added. On a related note, *in some cases*, it could be possible to decrease `sigma_max` instead of increasing `num_layers`, if the absorption stays strong enough while becoming more gradual due to the lower maximum value. I don't know if we want to go into that though.

tomflexcompute:
Thanks @momchil-flex these are great points. Indeed the reflection changes quite noticeably when resolution is changed, which can make high resolution simulations with absorber quite expensive. If one can define the thickness directly that would be great. I added a resolution test to the notebook and if at some point we added the thickness setting we can update it accordingly. 

Additional comments regarding your second point have been added to the final remark as well.

momchil-flex (approved):
Thanks!

PR #217, opened 2025-01-03
uploading EME bent waveguides example

FilipeFcp:
Hi all. Happy New Year!

The notebook was run using 2.8.0rc1.
Let me know your thoughts.


caseyflex (changes requested):
Thanks Filipe, this example is great and will be really helpful to users!

It would be nice to have a picture at the top of the notebook. Maybe both a picture of the type of physical structure being simulated (with bent waveguides), and a picture of the corresponding straight EME simulation.

In your first example of the 90 degree bend, I noticed you don't directly compare against FDTD. It might be nice to include this comparison. You could also show how the field plots differ so that users know that EME with bent waveguides can capture the radiation losses but the field visualization might be misleading (and some discussion along these lines).

I noticed some warnings in the solver log, would be good to check these. I also see some warnings regarding mode solver plane number of grid points. Does this actually cause a slowdown here? It seems like not an unreasonable size.

You should mention at the top that running the entire notebook costs more than 1 FC. In fact, the total cost is quite high, I wonder if there is any way to reduce it. On the other hand, it is really nice to see the direct comparisons between EME and FDTD, and the size of the FDTD simulations that we simulate here. Maybe you could split the notebook into two parts (within the same notebook, just with a clear separation between the two) -- the first part could be simple tutorial on EME bends with low FC cost, and the second part could be an advanced demonstration and comparison against FDTD with the larger structures fully simulated in FDTD.

Thanks again!

tomflexcompute (commented):
>  Also list the limitations of EME, for example I think you can't simulate an MZI with EME right?

I was thinking about this and indeed it wouldn't work for the usual directional coupler where two waveguides have opposite bends. On the other hand it can work for an MZI that doesn't have this but instead uses 50/50 MMI splitters. This generally made me think that we should also make another example to illustrate the [smatrix_in_basis](https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.EMESimulationData.html#tidy3d.EMESimulationData.smatrix_in_basis) method. @caseyflex didn't you have some scripts on simulating an MMI with EME?

FilipeFcp:
Thank you, @caseyflex and @tomflexcompute.
I have incorporated all your suggestions. For now, I’ve added the warning about the FC cost at the beginning and kept the FDTD and EME simulations together.

Regarding the warning messages, one arises due to gain in the modes because of the small plane size. I’ve added a note about that. I attempted to run with a smaller grid resolution to eliminate the other warnings, but the results, particularly when the losses are low, deviated too much from the FDTD ones.

caseyflex (approved):
Thanks @FilipeFcp. I think the warning even suggests making the mode plane larger in this case.

@tomflexcompute I guess we already have your nice MMI example with `smatrix_in_basis`:
https://www.flexcompute.com/tidy3d/examples/notebooks/MultiplexingMMI/
I was also testing EME on a version of this structure:
https://www.flexcompute.com/tidy3d/examples/notebooks/MMI1x4/
But, it was in the early stages of the solver -- I think it would be easier to start from scratch at this point if we wanted another example.

tomflexcompute:
For MMI it's pretty straightforward and EME is undoubtedly a great tool. I also have a EME vs FDTD benchmark on a 1x2 MMI that maybe we can also publish later to show the accuracy of EME, but yeah we have a (large) MMI example using EME already.

tomflexcompute (approved):

momchil-flex:
> @tomflexcompute I guess we already have your nice MMI example with `smatrix_in_basis`: https://www.flexcompute.com/tidy3d/examples/notebooks/MultiplexingMMI/ I was also testing EME on a version of this structure: https://www.flexcompute.com/tidy3d/examples/notebooks/MMI1x4/ But, it was in the early stages of the solver -- I think it would be easier to start from scratch at this point if we wanted another example.

Oh nice! I wasn't aware of that example. I just remembered your tests with the 1x4 one and noticed it was not in the main EME example. But yeah this is great.

momchil-flex (approved):

PR #218, opened 2025-01-06
uploading meep notebook

FilipeFcp:
Hi all. I am uploading the first version of the benchmark. This format is somewhat new to me, so I don't know if it is good. 

My initial thought is to exclude the Meep code and only upload the data. This way, users can run the benchmark without needing to install Meep. Additionally, we might prefer not to provide a tutorial on using Meep.

Also, I am not sure which rst file I should add the notebook.


tomflexcompute:
Thanks @FilipeFcp will take a deeper dive tomorrow but for the time being, the best place to place this notebook might be `docs/features/advanced.rst`?

alec-flexcompute (approved):
Hi Filipe, looks good! A couple small things:
1. The link in "application library" leads to a 404 error (also there should be a space after the comma)
2. The link in "example" one cell down also leads to a 404 error
3. Maybe specify which modifications you made to the example in the GUI?
4. I'm not sure where the Meep data is from. Maybe a comment on where you got the Meep data?
5. It may be helpful to plot the simulation, but also fine without it

tomflexcompute (commented):
Thanks @FilipeFcp that's great results. A few comments:

- Note again that h1 header only capitalizes the first letter of the first word and subsequent headers capitalize the first letter of all words. Here I also wonder if we can make the title a bit more informative like "Tidy3D FDTD accuracy benchmark" so it's more friendly for search engine. 

- In the intro we can add some more discussion about why we need to do this benchmark. Something along the line of "Tidy3D offers a lightning-fast FDTD solver capable of completing simulations in a small fraction of the time required by traditional FDTD solvers, reducing runtimes from hours or days to mere minutes. Importantly, we assure users that this enhanced speed does not compromise accuracy, as reliable results remain paramount to the utility of any simulation tool. "

-" application library,with those" -> " application library, with those"

- When I clicked the link to the MMI example in GUI, I got a 404 page. I think you might need to use this link instead https://tidy3d.simulation.cloud/workbench?taskId=pa-5e58d0d7-c3dc-446c-9309-fab8b94da423 since the original task was created by my account. 

- I don't have a clear quantitative idea but it seems like the inteference peaks in the tidy3d result and meep result are just shifted by a little bit in frequency. I wonder if it indicates numerically the device sizes are slightly different due to  resolution and subpixel. In any case I would suggest plotting the results in wavelength instead of frequency. 

- GUI exported codes often have rounding issues like 0.8999999999999999. They look a bit ugly so we can manually clean thoes up.

- We can add a `Final Remarks` at the end of the notebook to say something like "Tidy3D's accuracy has been systematically verified by our engineers interally as well as by users externally. If you have any question regarding the solver accuracy, we encourage you to perform a benchmark similar to what demonstrated in this notebook or [contact us](link to the support form page) for more information."

- In the last cell it seems like pandas is imported but not actually used?

- Also please black the notebook before merging.

FilipeFcp:
Thank you all.
I added this in the final remarks. Let me know if you have other suggestions:

"As we can see, the results deviate only due to numerical errors, likely caused by differences in meshing and subpixel averaging strategies.

Additionally, one could argue that the speed comparison may not be entirely fair, as the Tidy3D simulation was run on more powerful hardware. However, a direct comparison is not possible since Meep does not support GPU acceleration.

From a user’s perspective, both simulations were built and submitted on the same laptop. Therefore, Tidy3D offers the advantage of cloud computing, meaning that users do not need powerful hardware on hand to benefit from the speed advantage, nor do they need to dedicate a workstation for hours to a single simulation."

tomflexcompute (approved):
Looks very good @FilipeFcp thanks!

tomflexcompute:
Tagging @momchil-flex to take an additional look here. In terms legality, it should be totally fine to mention keep right?

momchil-flex:
Thanks for this @FilipeFcp !

Regarding legality, I think there is no issue, as meep is not commercial. Also might be important to note that our notebooks repository has AGPL-3 license, which is consistent with meep's GPL license (the distribution of GPL-licensed code, even when just using the provided api of a library, as non-GPL-licensed code is always a grey area).

One question, since meep uses uniform grid, why don't you try using that in Tidy3D too? Setting dl to `1 / 30` as in meep. Maybe this will make the agreement better (although to really make them agree one would probably need to make sure that the grids match 100% even within that setting, as there could be some numerical reasons why they still don't.)

There's also a bunch of warnings - do you know why? Would be nice to not have them in the docs.

```
WARNING:urllib3.connectionpool:Connection pool is full, discarding connection: flow360studio.s3.us-gov-west-1.amazonaws.com. Connection pool size: 10
```

tomflexcompute:
This Connection pool size has been very annoying and @daquintero will look into it soon.

FilipeFcp:
Yes, I tried the uniform grid first, but the results are not as good as with the AutoGrid. It looks like this (dl = 1/30):
image

Playing a little, we can have a good agreement with dl = 1/28:

image

I tried the best I can to match the Meep simulation size with Tidy3D, but there is a difference in the PML positioning. We position outside the simulation domain, Meep inside. 

Anyway, I thought that ultimately the subpixel averaging strategy might dominate the differences, does that make sense?

tomflexcompute:
Also just curious if the meep run time reported in the notebook is still running on a single-core or multi-core @FilipeFcp ? If it's single core it might not be a super fair comparison I feel like and if so we can maybe remove the exact number and just say something generic like "hundreds of minutes".

e-g-melo (approved):
Thanks @FilipeFcp! It is very exciting to see these benchmark simulations.

- Include a `_ = sim_base.plot(z=0)` on cell [8] to suppress the output.

The final remarks are very nice!

tomflexcompute:
@FilipeFcp actually before merging, could you update https://github.com/flexcompute/tidy3d-notebooks/blob/develop/misc/import_file_mapping.json since you import the meep result from the misc folder?

FilipeFcp:
Hi all. I was waiting to merge because I wanted to run a test using more processors. However, to do that, I needed to close everything on my computer; otherwise, it would crash due to lack of memory. Anyway, I ran the test this weekend, and the simulation time was reduced from 830 minutes (1 core) to 690 minutes (10 cores). It is less than what I was expecting, but I checked, and all 10 cores were being used. I’ve added this information to the notebook. I also updated the import_file_mapping and added the plt.show as Emerson suggested.

tomflexcompute:
Looks good to go. Thanks @FilipeFcp for the great effort!

PR #220, opened 2025-01-08
Alec/var eps fix

alec-flexcompute:
Hi guys, updating the variational effective index approximator due to an error in the previous version.


tomflexcompute (commented):
Thanks @alec-flexcompute for the fix. 

- Note that headers besides the title have to be h2 (##) or higher otherwise the docs build will be messed up since it automatically recognizes all the h1 headers as titles. For example: "# Example 1: 2D Planar Waveguide Bragg Grating" needs to be changed to ## or higher.  

- In the examples, could you add some simple comments in the code?

- "Strip-to-slot counter-example" -> "Strip-to-slot Counter-example"

FilipeFcp:
Thanks @alec-flexcompute, that’s a very nice tutorial! I remember when I was working with Nanobeans, I first tried some 2D simulations to get at least close to the intended resonance, but my approach was much more naive.

Just a few comments:

- It might be a LaTeX rendering issue on my end, but I can't see equations in the markdown below cell [1]. It works if I replace the `</center><be>` with just `$$`.
  
- In the same text, if I understood correctly, `x` is the extrusion dimension and `z` is the waveguide propagation. This is not necessary, but maybe a quick clarification would be helpful.

- In the comments of cell [9], I believe "3D" should be "2D" and "2D" should be "1D."

alec-flexcompute:
Thanks for the great catches @tomflexcompute and @FilipeFcp ! I've implemented the changes.

FilipeFcp (approved):

tomflexcompute:
Thanks @alec-flexcompute looks ready to merge.

tomflexcompute (approved):

PR #221, opened 2025-01-09
uploading radar absorbing metamaterial notebook

FilipeFcp:
Hi all. I am uploading a case that we modeled last year for a user. The results match the paper very well, so Tom said it could be a good showcase. Also, it seems a good case to add to our microwave examples.


tomflexcompute (commented):
Thanks @FilipeFcp this is great. 

- Could you mention the length unit by default is um and frequency unit is Hz? Some users might be unfamiliar with our unit system and get confused by the numbers you defined in cell[2]. Also could you add brief comments on what these geometry parameters correspond to? 

- In the resistivity calculation, could you also mention its unit, comment on where "60" isfrom and what it means?

- It seems like only normal incidence is simulated. Is it necessary to use Bloch boundary instead of periodic? Since it's broadband simulation, if oblique incidence is needed, using the constant angle plane wave source is more recommended. 

- Before merging, could you rerun the notebook so all cells are run in order? 

- In the metadata, `applications` and `features` fields are missing. For applications, this probably can be both a metamaterial and a RF application.

- "metametarial" -> "metamaterial"

- "smaller then" -> "smaller than"

alec-flexcompute (approved):
Hi Filipe, great work as usual! I especially liked the elegantly written code in the windmill definition :)

- “we will reproduce the results publish” -> “we will reproduce the results published”

- For the normal incidence comment of Tom’s, maybe just specify that if you wanted to modify the simulation to inject at an angle, you’d use these boundary conditions (I think it’s good that you included this info anyway in case people wanted to modify this simulation for their own applications). I just think the current section may confuse some people with the data corresponding to a non-angled source.

Other than that and Tom's comments, I think it looks great

FilipeFcp:
Thank you @Tom and @alec-flexcompute!
I implemented everything.

tomflexcompute:
Thanks @FilipeFcp all look good. I just don't see any metadata after the update. Is it my viewer issue?

image

Also in the introduction it still mentions using Bloch but I think you can remove it since you have switched to periodic.

e-g-melo (approved):
Hi @FilipeFcp! Thanks for another high-quality example!

All looks great! I only suggest defining the PMMA refractive index as a variable, but that is not critical.

tomflexcompute:
@FilipeFcp I still don't see the metadata from my side for some reason. @e-g-melo do you see it correctly?

tomflexcompute:
Actually for some reason, I can see two branches with the same name. One is updated and the other one is not.

In the updated one, the `applications` and `features` fields are still missing. Is that correct?
image

FilipeFcp:
It should be fine now

tomflexcompute:
Thanks @FilipeFcp good to merge.

PR #223, opened 2025-01-17
fix[notebooks]: fix autograd level set notebook

### LINKED ISSUE #205: Fix level set notebook for autograd
`Autograd10YBranchLevelSet.ipynb` currently doesn't work and needs to be fixed. Biggest issue is probably aux data handling.

groberts-flex:
This PR updates and fixes the autograd level set notebook so that it matches the optimization happening in the adjoint version (with Jax). The major changes were to (1) remove references to the Jax backend and replace with autograd, (2) fix parameters that were set incorrectly, (3) match function calls more correctly between Jax and autograd (i.e. - the np.gradient with a certain mesh size), (4) and fix an underlying bug in the tidy3d codebase to get gradient magnitudes correct in the CustomMedium class (note that an update to the server python code needs to be in effect for this code to run as it does locally).


yaugenst-flex (changes requested):
Hey @groberts-flex thanks this looks great!

Just very minor comments:
 - There are some debugging comments I think left in the notebook (`# does optax convert it to a jax array?`)
 - `from tidy3d.plugins.autograd.differential_operators import value_and_grad` should work as a top level import so `from tidy3d.plugins.autograd import value_and_grad` would be preferred
 - While we're at it, could you run `ruff format` on the notebook? We should start doing this for every notebook.

groberts-flex:
great, thank you for these comments! think I got all the debugging comments out of there now, fixed the import, and ran ruff format on the notebook file!

yaugenst-flex (approved):
Nice!

PR #225, opened 2025-01-20
Added the PBS notebook

tomflexcompute:
- Added the PBS notebook and relevant materials
- Fixed the title capitalization of the prefab inverse design notebook. 

Hi guys thanks in advance for reviewing. This postprocessing and plotting part is a little bit annoying and I tried to avoid code repetition but it still looks a bit more complex. Hope that's not very difficult to understand. I purchased access to the referenced paper. If you would like to take a look, please let me know.


alec-flexcompute (commented):
**for PBS notebook:**

Looks great, Tom! The only thing that might cause confusion is how complicated this structure is. Perhaps a bit more clarity concerning the design might be useful. For example

- introduce what the "duty cycle" is or clarify using the diagram variables.

- describe at the beginning what parameters are being optimized (maybe near the diagram?) in this notebook

As for the results and style of this notebook, it looks great!

**Fab Aware notebook:**

- Above cell 6: "Then we set up our..." -> "Here we set up our..."

- It also might be useful to include a diagram of the simulation here, but since there's a plot of the simulation a couple cells down, it would not be super necessary

- Above cell 14, it's not quite clear what you mean by "match this resolution" or what a "buffer layer" specifically entails

- Above cell 19, maybe more specific description of overall metric

- two blank cells at the end :)

Once again, impressive job @Tom!

FilipeFcp (approved):
Hi @tomflexcompute. Very nice! I have nothing to add to Alec's comments.

e-g-melo (approved):
Hi @tomflexcompute! Thanks for the example. It is really cool.

I only suggest including something about the `n_010_example` meaning.

Emerson

tomflexcompute:
Thanks all. I think @alec-flexcompute you tagged the wrong person.

PR #227, opened 2025-01-21
Rework conversion between autograd/jax arrays to avoid xla error

yaugenst-flex:
Replaces `params = optax.update(params)` with `params[:] = optax.update(params)` in all relevant notebooks, this gets rid of any type conversion errors in `params`. It should also partially fix some memory issues.

I know we have discussed previously that in-place assignment might be unfamiliar to some users, but I think it is the correct way to do this. It's not any less confusing than scattering `np.array()` conversions throughout the notebooks, so I think it's ok if we are consistent. Once we start adding some non-optax notebooks, this will be a non-issue anyway.

@e-g-melo this fixes [SCEM-8965](https://flow360.atlassian.net/browse/SCEM-8965)

Also includes a bunch of formatting changes for consistency (ran ruff on the modified notebooks)


e-g-melo:
Hi @yaugenst-flex! Thanks for looking into this!

The Jax-based notebooks are still available in our example library. Should we also update them?

Do we have plans to remove those notebooks from the example library?

tomflexcompute (approved):
Thanks @yaugenst-flex . Did you rerun these notebooks btw?

yaugenst-flex:
@tomflexcompute yes i reran all notebooks and they all work :)

yaugenst-flex:
>The Jax-based notebooks are still available in our example library. Should we also update them?

The jax notebooks don't have this conversion issue so for now it's fine I think.

> Do we have plans to remove those notebooks from the example library?

We are deprecating the jax plugin but probably we want to keep these notebook around for a while. We should probably organize them better though...

PR #233, opened 2025-01-23
Added charge validation notebook

marc-flex:
Adding Charge validation notebook

for visibility: @momchil-flex @dbochkov-flexcompute


momchil-flex:
I know it's still WIP but a couple quick comments:

- [x] Could you try to clear warnings as much as possible? Regarding this one
image

it should go away if you make the mode plane either a bit smaller or a bit larger. Not sure if we need to rethink this altogether, but anyway it may go away automatically for the ModeSolver when we convert it to a ModeSimulation that no longer contains and FDTD Simulation inside of it.

- [x] We have ModeSolver, EMESolver, and HeatSolver notebooks that discuss the basics of these, respectively. Shall we just call this one ChargeSolver? And the title can be either just `Charge solver` or `Using the charge solver to simulate a phase modulator`.

daquinteroflex:
Currently working on as much decoration/polishing as I can too on this notebook btw as talking with Marc.

marc-flex:
Thanks @daquinteroflex for taking care of this!
I know this is still WIP but I had a quick look and I have a couple comments:

- Nothing against having the equations here as well but I thought it'd be sufficient to redirect people to the docs?
- [x] In the line `(constant temperature) case. This simulation performed at T=300. ` Kelvin units are missing
- [x] The current wording seems to imply that the potential is the solution to Gauss' equation and n and p the solution to continuity equations, which of course, isn't true. The system is fully coupled. So maybe a wording along the lines of `Electric potential, n and p are computed by solving the following coupled system of PDEs...`. What do you think?

daquinteroflex:
- Nice yeah agree with 2, 3
- So I understand our notebooks are nominally "the main reference" and we've added equations before for multiple notebooks eg. [heat](https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/HeatSolver.html), [microwave](https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/CharacteristicImpedanceCalculator.html#Initialize-monitors-and-finalize-simulation), [time modulation](https://docs.flexcompute.com/projects/tidy3d/en/latest/notebooks/TimeModulationTutorial.html#More-Complicated-Modulations) ... So maybe we want to keep doing this, maybe especially for our first charge solver simulations maybe makes sense to explain it for the uninitated?

momchil-flex:
Some other small things:
- [x] The aux_left and aux_right media are identical, I think this may be a remnant from a previous version in which the media were used to define the BCs instead of the structures, which is now the case? So a single medium can be used (or if keeping them separate then maybe explain use cases in which we might need that?) Also, `conductivity=1` is pretty arbitrary I assume You'd expect the same results with 0.1 or 10? Maybe a comment on that.
- [x] `air` medium is still defined with a `heat_spec`, which technically we're deprecating? Should it also just be a MultiPhysicsMedium?
- [x] `Once we have out doping` -> `our doping`
- [x] `In the plots below, we have use` -> `we have used`
- [x] `and positive voltages to the n side` -> `and a positive voltage`
- [x] `this class a member convergence_dv` ->  `this class takes an argument covergence_dv`
- [x] `we want to solve for 0V and 20 bias` -> `20V`
- [x] A few instances of "as it can be seen" - I think just "as can be seen" is the standard way of saying it

marc-flex:
> The aux_left and aux_right media are identical, I think this may be a remnant from a previous version in which the media were used to define the BCs instead of the structures, which is now the case? So a single medium can be used (or if keeping them separate then maybe explain use cases in which we might need that?) Also, conductivity=1 is pretty arbitrary I assume You'd expect the same results with 0.1 or 10? Maybe a comment on that.

That's right. We only need 1 material. The conductivity value is arbitrary since Charge does not currently solve for conductors

momchil-flex:
Sorry, I clicked comment too early and have been updating my comment above with the other small things I found when reading through. :)

daquinteroflex:
So I've only got removing the warnings left but apart from that would say it's nearly there to do another review pass

daquinteroflex:
> it should go away if you make the mode plane either a bit smaller or a bit larger. Not sure if we need to rethink this altogether, but anyway it may go away automatically for the ModeSolver when we convert it to a ModeSimulation that no longer contains and FDTD Simulation inside of it.

I'm not sure the mode plane size sorts it as I've tried both edges. Still getting more familiar with how all validators are currently fully setup now but maybe @yaugenst-flex can help in the interim much faster than me?

daquinteroflex:
Note that this might need to be merged into this branch before merging to pre/2.8 https://github.com/flexcompute/tidy3d-notebooks/pull/235

docs/features/charge.rst
```diff
@@ -0,0 +1,10 @@
+Charge |:zap:|
+~~~~~~~~~~~~~~~~~~
+
+This section introduces the Charge solver. The examples demonstrate basic usage, boundary conditions, and material setup.
+
+.. toctree::
+    :class: example-notebook-toc
+    :maxdepth: 1
+
+    ../../ChargeSolver
```
daquinteroflex, line 10: Fyi @marc-flex for the other notebook

daquinteroflex:
Shall we squash and merge this?

Issue #231, opened 2025-01-23
Replace tidy3d.plugins.mode.web::run_batch with the native web.Batch


momchil-flex:
Many notebooks have been converted to use the `run_batch` function that was specifically implemented for the mode solver. However, this turns out to be unstable in some cases as it can e.g. open too many connections and time out. Now that the native `web.Batch` is enabled for the mode solver in 2.8, we should convert the examples to use that (and make sure all works fine).

@tomflexcompute who should take this?

tomflexcompute:
Yeah good to switch. I'll modify the notebooks and push a PR. I think there are only a few notebooks using run_batch right? Are you aware of any?

momchil-flex:
I noticed that quite a lot turned up in this search which is why I made this issue. I assume at some point someone incorporated it, maybe @prashkh .

https://docs.flexcompute.com/projects/tidy3d/en/latest/search.html?q=mode+run_batch

tomflexcompute:
Oh yeah it's more than I thought. I'll change it. Do we have a timeline for 2.8.0 release?

momchil-flex:
Will be wrapping 2.8.0rc2 up soon but then we want to let it get well tested before official 2.8.0, so not soon. Maybe end of February.

tomflexcompute:
Sounds good most people will still be in 2.7 for a while and I can update it before the 2.8 release.

PR #234, opened 2025-01-23
Updated notebooks to use regular web api for batch mode solver

tomflexcompute:
Happened to find some time today so decided to address this PR https://github.com/flexcompute/tidy3d-notebooks/issues/231#issuecomment-2610149901 sooner rather than later. Hope it doesn't cause conflicts when merging to develop later.

There are two charge solver notebooks that I haven't modified yet, but they will be revamped soon to use our own charge solver right? They can be modified together. There is also a heat solver notebook that I haven't modified either. Will that notebook be modified due to some API change in 2.8.0 or I should just modify it now?


momchil-flex:
Thanks @tomflexcompute! I asked @marc-flex to handle the substitution in the heat/charge notebooks. There's also one notebook which is currently only on develop which uses this (the radiative loss analysis). We can fix this too eventually after merging develop in pre/2.8 (for now we're not doing that because of the docs build issue.. too many notebooks :D )

tomflexcompute:
No problem I can fix that later on.

e-g-melo (approved):
Hi @tomflexcompute! Thanks for updating these notebooks. A single approach for running batches is a great improvement in usability!

tomflexcompute:
I suppose at some point we need to update all the single mode solver runs to use `web.run()`, which requires updating many more notebooks but that's not very critical at he moment so we can update them closer to the 2.8 release. Shall I merge this PR as it is @momchil-flex ?

marc-flex:
I was going to add here a modification to the MZ interferometer, but I can I move it to a different branch :+1:

tomflexcompute:
> I was going to add here a modification to the MZ interferometer, but I can I move it to a different branch 👍

Oh I see please feel free to go ahead and update it here. Whatever is more convenient is fine.

momchil-flex:
I discussed with @marc-flex that he'll keep the heat/charge notebooks in a separate PR as there are other updates too.

PR #238, opened 2025-01-28
:cherries: Cherry pick fixes from dario/charge

daquinteroflex:


daquinteroflex:
So this needs to be done in order, Daniils first and then mine

PR #237, opened 2025-01-28
:zap: :fire: Update old heat-charge notebooks

daquinteroflex:

PR #239, opened 2025-01-28
:cherries: :fire: :zap: Combined set of fixes and notebook upgrades for Charge

daquinteroflex:
Combines and closes https://github.com/flexcompute/tidy3d-notebooks/pull/237 and https://github.com/flexcompute/tidy3d-notebooks/pull/237


ChargeSolver.ipynb
marc-flex, line 1: Since we're now extending the structures to inf in the z direction we need to revisit the wording. I'll create an issue for this

HeatSolver.ipynb
marc-flex, line 1: here we have changed the simulation object to `HeatChargeSimulation`. We could also change `FluidSpec` and `SolidSpec` to `FluidMedium` and `SolidMedium`

ThermallyTunedRingResonator.ipynb
marc-flex, line 1: Similar to the above

marc-flex (approved):
Leaving the review for reference. This can be fixed later on on a related issue.

PR #248, opened 2025-02-13
Added the PEC sphere benchmark model

tomflexcompute:
This PR adds a monostatic RCS of a PEC sphere benchmark. 

A new `Accuracy Benchmarks` section is created and a few notebooks are moved to this section including @FilipeFcp 's recent meep benchmark. This way users who have questions about accuracy can find all the benchmarks in one place.


alec-flexcompute (approved):
Amazing work Tom! I literally only have two minor nitpicks:
1. Maybe a couple words on _why_ TFSF is the best source for this calculation, as mentioned above cell 3
2. The final blank cell has a line break in it

Excellent agreement with the analytical result!

FilipeFcp (approved):
That is a really nice example. The agreement is great! I have no further comments.

weiliangjin2021 (approved):
All look great!

PR #249, opened 2025-02-18
Added the mode overlap tutorial

tomflexcompute:
This notebook demonstrates the recently added `GaussianBeamProfile` class and `translated_copy` method to calculate the mode overlap integral. 

Note that the links to the API reference are not live obviously but the links should be correct.


momchil-flex:
This is awesome, thanks! Some people have been asking for a while...
Just some minor comments:
- The plots are transposed, x is on the y axis and vice versa. One way to fix that is to pass `x="x", y="y"` to the field `.plot` method.
- You should comment on the fact that the GaussianBeamProfile expects one of the `size` values to be 0, and takes that as the normal axis w.r.t. which the propagation (and the angle_theta, angle_phi) is defied.

tomflexcompute:
Thanks @momchil-flex ! Good points.

FilipeFcp:
That is a really great example of how it can be done simply with just a few lines of code.
My only comment is right above cell [9]—I believe "Gaussian bean" should be "Gaussian beam."

FilipeFcp (approved):
That is a really great example of how it can be done simply with just a few lines of code.
My only comment is right above cell [9]—I believe "Gaussian bean" should be "Gaussian beam."

tomflexcompute:
> That is a really great example of how it can be done simply with just a few lines of code. My only comment is right above cell [9]—I believe "Gaussian bean" should be "Gaussian beam."

oh haha I think I was cooking black bean soup when writing this.

tomflexcompute:
Should be good to go I think @momchil-flex a quick final look?

momchil-flex (approved):

PR #251, opened 2025-02-25
Updated MZI to use Charge

marc-flex:
Updated the Mach-Zehnder Interferometer notebook to use Charge. 

Due to the difficulties I found in converging to a solution at $V_\pi$ I had to increase the voltage step plus I had to add an additional voltage. Will look further into that in the back-end.

fyi @tomflexcompute


tomflexcompute:
Thanks for the update @marc-flex . Should it be merged into develop so it makes into 2.8.0? Some minor comments below:

1. The devsim imports can probably be removed right?

2. Could you link the classes in the comments like `SemiconductorMedium`, `ConstantDoping` to the API docs for easier reference?

3. In cell[11] can you add plt.show() to get rid of the axes printout? 

Adding @FilipeFcp here to take a look too since he was experimenting with something similar and ran into some convergence issues too.

marc-flex:
Thanks @tomflexcompute 
I have addressed your comments. For some reason we're missing an entry in the docs for `IsothermalSteadyChargeDCAnalysis`. I have already asked @daquinteroflex. As soon as we get that I will update.

marc-flex:
Also, once all comments are in I'll rerun the notebook so that it's clean

alec-flexcompute (commented):
Hi Marc, excellent job! The only substantive feedback I'd suggest in addition to Tom's is a brief note on why the absolute tolerance value is what it is, and why this is acceptable, as new users might be worried about such a seemingly large value, as well as a brief note on why the convergence step was chosen. Other than that,

- above cell 3, "variable" -> "variables"
- above cell 12, "VoltageBC can accpet an array of voltages as source" -> "accept an array of voltages as sources"
- above cell 24, "so that we assing" -> "so that we are accessing"
- above cell 34, capitalize
- above cell 40, "dividinng" -> "dividing", "obtaining S-matrix" -> "obtaining the S-matrix", "perform simulation" -> "perform the simulation"

Nice example!

FilipeFcp (approved):
Hi Mark,

The updated notebook is really nice!

A few comments:

- In the introduction, it would be nice to have a link to some Charge documentation.
- Below cell [1], I believe spitter → splitter.
- Below cell [3], semiconductror → semiconductor.
- Maybe delete the commented Si import in cell [6]
- Maybe add a few comments regarding `rel_tol` and `abs_tol`.
- I think we can remove the # devsim setting  comment in cell [16]
- I think the web import in cell [19] is not necessary.

Also, I have some general comments and random questions:

- In cell [11], why do we need two auxiliary materials with the same properties but different names?
- When creating the charge simulation, you define a small buffer `size=(0, 2*x_total+1, h_side+1)`. Would it be a problem to remove that `+1` and let the simulation boundaries touch the structure?
- If we could have a built-in function for Si to simplify cells [21, 22, and 23], that would be cool.

My simulation was not working because I was using a wrong simulation size (for some reason, I inserted it as half the Scene size). In this case, the simulation errored during runtime without a clear error message.

alec-flexcompute:
Hi Marc,

Also, this variable is used inside the junction-creating function but is defined outside of it, and isn't a parameter. Might want to change this
image

marc-flex:
> above cell 12, "VoltageBC can accpet an array of voltages as source" -> "accept an array of voltages as sources"

@alec-flexcompute wouldn't that imply that there are many sources instead of just one taking up different values?

marc-flex:
>  In cell [11], why do we need two auxiliary materials with the same properties but different names?

I changed it to have only 1.

>  When creating the charge simulation, you define a small buffer size=(0, 2*x_total+1, h_side+1). Would it be a problem to remove that +1 and let the simulation boundaries touch the structure?

It shouldn't be a problem. I'll adjust to include pin + contacts only

> If we could have a built-in function for Si to simplify cells [21, 22, and 23], that would be cool.

[This issue](https://github.com/flexcompute/tidy3d-core/issues/846) will address this.

momchil-flex (approved):
Awesome @marc-flex, things are getting cleaner!

Instead of `run_mode_batch` could you switch to the native `web.Batch` which we now support for the mode solver too, and set `reduce_simulation=True` in `Batch.run` so the warnings will be avoided.

> > above cell 12, "VoltageBC can accpet an array of voltages as source" -> "accept an array of voltages as sources"

> @alec-flexcompute wouldn't that imply that there are many sources instead of just one taking up different values?

I guess to avoid confusion let's just add a sentence afterwards like

```
In Charge this can readily be done since the VoltageBC can accpet an array of voltages as source through DCVoltageSource. A parameter scan will be run and the returned data will have the provided voltage values as a separate dimension.
```

marc-flex:
All comments have been addressed. If you're all happy with the current status I'll merge in the morning.

PR #252, opened 2025-02-26
Rerunning structural color notebook with broadband plane wave

momchil-flex:
We finally fixed the flux of the plane wave source when it is very broadband. Now the flux values in a few cases still go a tiny bit above 1, but it is on the scale of 1e-4.

Any other example that may profit from a rerun?

By thew way @e-g-melo this table looks weird, no..? Not due to my PR. How should it be fixed?

image


tomflexcompute:
That's great! As far as I can tell, this is the only example that has visibly larger than 1 flux.

e-g-melo:
Thanks @momchil-flex! It seems very good!

Maybe format the header labels to add more information. For example, Wavelength (um), period = 0.3 um, period = 0.31 um, period = 0.32 um, and so on. Or ideally, add only one "period (um)" label on the left to make it cleaner. I can try it if you want.

momchil-flex:
But something is wrong with the wavelength values no?

momchil-flex:
Oh sorry I think I'm misunderstanding this. It's just the header of the table. I need to stop for today. :)

momchil-flex:
Tried a little but it seems annoying to fix, so just leaving as is.

PR #259, opened 2025-02-28
Minor docs fixes

tomflexcompute:
Minor docs metadata fixes to ensure the examples go to the correct category. 

In the plasmonic nanoparticle notebook, I tried to increase the resolution to match the result better. Forward scattering matches well but total absorbed power and total scattered power don't seem to converge to the analytical results. @weiliangjin2021 any idea why and how to improve it?

### DIFF
--- a/AnisotropicMetamaterialBroadbandPBS.ipynb
+++ b/AnisotropicMetamaterialBroadbandPBS.ipynb
@@ -1296,8 +1296,7 @@
   "feature_image": "./img/pbs_swg_am.png",
   "features": [
    "Mode analysis",
-   "Parameter sweep",
-   "Anisotropic material"
+   "Parameter sweep"
   ],
   "kernelspec": {
    "display_name": "Python 3 (ipykernel)",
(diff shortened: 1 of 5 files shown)


e-g-melo:
@tomflexcompute, have you removed or updated the applications and features metadata?

image

tomflexcompute:
That's correct. Sorry I should've clarified earlier. I'm moving the notebooks with comparison to analytical solutions or other solvers to the new `Accuracy Benchmarks` section in Python tutorial instead of the example library so users can find all the accuracy benchmarks together, which is a common question.
image

e-g-melo (approved):
Got it! Thanks!

weiliangjin2021:
> In the plasmonic nanoparticle notebook, I tried to increase the resolution to match the result better. Forward scattering matches well but total absorbed power and total scattered power don't seem to converge to the analytical results. @weiliangjin2021 any idea why and how to improve it?

Probably related to staircasing, that is the default option for metallic interface. I tried volumetric averaging, but it doesn't help. Could you make a sub-issue [here](https://github.com/flexcompute/tidy3d-core/issues/833)?

PR #263, opened 2025-03-04
SWG Waveguide Crossing model

yuanshen-flexcompute:
A new model of a subwavelength grating (SWG) multimode waveguide crossing


tomflexcompute (commented):
Thanks @yuanshen-flexcompute for the great first notebook. Great quality and details are well thought through. A few minor comments:

1. We follow a strange convention where in the notebook title, only the first letter of the first word is capitalized. In subsequent headers, the first letter of each word is capitalized. Could you fix the capitalization of the notebook title? 

2. The schematic images are very helpful, but they appear to be a little to large. Could you reduce their size by about half? Showing them at the current size also makes them look a little blurry.

3. After the DOI in the introduction paragraph, it's missing a period. 

4. Usually setting the logging level to error is not recommended unless there is really no way around it. In this case I think the warning is mainly due to structures too close to pml? If so, we can add some space to get rid of the warning. 

5. In the boundary condition part, maybe add a sentence on why we increase the number of layers in the x direction. 

6. Currently the `plot_3d()` line is commented out. You can actually run this cell and the 3D plot will be available for users to see when the notebook is built into our learning center, as the example [here](https://www.flexcompute.com/tidy3d/examples/notebooks/YJunction/#Simulation-Setup) shows. 

7. Add the schematic image as the cover image for the learning center after you create it.

alec-flexcompute (approved):
Great job Yuanshen, a very cool example! A couple things:

1. (Optional) Put the intro before the cell where you import the libraries and change the config level.

2. In the "Extending this to 2D" paragraph, it's scrollable. Can you make it not scrollable?

3. (light suggestion) Set the final mode transmission plots on the same vertical axis to visually highlight the difference between through and cross

4. Rerun so the cells are run in sequential order (the numbers next to the cells should be sequential)

FilipeFcp (approved):
Hi @yuanshen-flexcompute,

This is a very nice and informative notebook!

My only comment (optional) is to clarify that we are not using PML since they can cause divergence with the dispersive material. You might also consider adding [this](https://docs.flexcompute.com/projects/tidy3d/en/latest/faq/docs/faq/Why-is-a-simulation-diverging.html) link. 

Additionally, it seems that the cell with the 3D plot wasn't run.

tomflexcompute:
Thanks @yuanshen-flexcompute for the revision. All looks great. 

In the features you added "mode analysis", which refers to using the mode solver to analyze the waveguide mode. In this notebook it doesn't seem like the case. Maybe you can remove it. 

The 3d schematic image can be added right after the first introduction paragraph. I think it looks very nice. 

Everything else looks good to go!

tomflexcompute (approved):
I think it's good to go! After you change the local git config, would it be fine to revert the changes in the .gitignore file? If so you can revert the changes here. 

Please squash all commits in this PR and then rebase/merge. Thanks @yuanshen-flexcompute !

PR #264, opened 2025-03-06
add notebook demonstrating Bayesian optimization of PRUC

dmarek-flex:
notebook demonstrating steps in the AP-S/URSI submission


weiliangjin2021:
Looks good. If you have time, probably good to replace some of mesh override structures and all snapping points with layer refinements.

tomflexcompute (approved):
Thanks @dmarek-flex could you update the rst file as well? I'll draw a 3d schematic image for you shortly.

dmarek-flex:
> If you have time, probably good to replace some of mesh override structures and all snapping points with layer refinements.

Yea, definitely! I was going to try that but ran into the problem trying to reproduce results with the new solver version! We will definitely revisit this notebook with those changes and more.

PR #270, opened 2025-03-17
Add first draft of notebook

alec-flexcompute:


yuanshen-flexcompute (commented):
Really cool work! It's great that there's such good match with the paper results. I like the flow of the notebook overall. The comments below generally pertain to adding more written explanation within each section, which helps to guide the reader who might be unfamiliar with the code and application. 

- It might be helpful in the opening section to explain why it's important to consider heat dissipation in photonics applications, just to provide some context and motivation. Even if the paper goes into detail, I usually assume the reader is either too lazy/busy to read it, or may not even have access to it. 
- Cell  1: "1. A heating element [at] a fixed height..."
- Section "Create Structures....": Consider plotting the scene cross sections with the heating element at various lateral displacements. Not all, but 2 or 3 could help to reinforce the intent of the simulation. Also could include a sentence or two explaining it. 
- Section "Other Heat Simulation Specifications": Maybe a few sentences explaining the intent of each cell? 
- Section "Create Batch...": It looks like symmetry is only used for the zero lateral displacement case. Perhaps include a sentence or two explaining that? In the plot of the simulation cross section, might also be worth pointing out the temperature monitor and the heat source.
- Section "Get and Compare Data": Perhaps include a couple of sentences discussing the results?
- Section "Second Setup...": Points below.
- In the first code block, the portion that defines the stack geometry could use a few more comments describing the geometry operations. If it is not too much work, it might even be helpful to include a simple sketch with the relevant dimensions labelled.
- I wonder if the make_stack() function can be split and/or pared down a little bit. The BC and grid spec, for example, can re-use previous definitions. Other static definitions could also be pulled out into a separate code cell.
- Section "Plot and Compare Data": In the first plot, the markers for the paper data are a bit hard to read, could be bigger or a different shape. It would also be helpful to have a couple of sentences describing the contents of each of the plots. 
- Since this is a rather long notebook, it might be helpful to have a "Conclusion" section that summarizes everything in a few sentences, re-iterate why it's meaningful, and perhaps point the reader to other Heat example models. 

Again, great work!

tomflexcompute (commented):
Thanks @alec-flexcompute for the fantastic results. Great validation of the solver accuracy. Some comments mostly related to formatting:

1. Notebook title needs to follow other notebooks, i.e. something like `HeatDissipationSOI.ipynb`.
2. Missing a period at the end of the first paragraph. 
3. "A heating element a fixed height above" -> "A heating element with a fixed height above"
4. SiO2 can be formatted as SiO$_2$ to make 2 a subscript. 
5. Capitalization of the first word in the comments are a bit inconsistent. It's not capitalized in the beginning but is later on. 
6. "Offset vs wWaveguide" -> ""Offset vs Waveguide"
7. "Heater offset" -> "Heater offset (μm)" and "Waveguide temperature change (°C)"
8. The imported data files need to be added to the misc folder
9. The notebook needs to be added to `docs/case_studies/pic_active.rst`
10. I didn't check since I'm reviewing directly in github but make sure the metadata are added if not already.
11. Not really relevant to the notebook but for future PRs ideally the title of the PR can be more specific like "Adding the first draft of the heat example notebook" so it's easier to locate in case in the future we need to look for this PR for some reason. 

I'll prepare a schematic image for you to use later today.

alec-flexcompute:
Thanks for the feedback, I'll implement and push when the schematic comes in!

alec-flexcompute:
Changes implemented

yuanshen-flexcompute (commented):
Nice! Just some minor comments:

* In the first cell, last line the "2" in "SiO_2" is split into a separate line. Perhaps you can typeset it as "$\text{SiO}_2$"?
* In section "Get and Compare Data", looking at the results plot, what is the difference between "Experiment" and "Pant et al."? Is the latter a simulation result?
* In the later section "Plot and Compare Data", looking at the first plot, in the legends the labels are "Tidy3D 0", "Paper 0" etc... Perhaps clarify that the number refers to the arm displacements in microns?
* This might be my mistake, but I am still seeing the older notebook "heat_dissipation_soi.ipynb" along with this new one. Should that be deleted?

tomflexcompute:
Thanks @alec-flexcompute besides the comments from Yuanshen above. The [metadata](https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e) is missing. Other parts look very good.

alec-flexcompute:
Thanks for the feedback! Changes should be implemented

yuanshen-flexcompute (approved):

tomflexcompute:
Thanks @alec-flexcompute I applied some minor edits on the metadata and should be good to go. Please squash the commits before you rebase/merge.

tomflexcompute (approved):

PR #271, opened 2025-03-21
Minor fixes

tomflexcompute:
- Fixed the Bayesian optimization notebook title to h1 header (it was incorrectly set as h2 header). Fixed the thumbnail image file name.
- Minor modification of the SWG waveguide crossing metadata and thumbnail image size.


FilipeFcp (approved):
Good catch. Seems all good!

PR #272, opened 2025-04-09
Simplified the quick start notebooks

tomflexcompute:
Simplified the quick start notebooks


yaugenst-flex (approved):
Thanks @tomflexcompute, these look great, much simpler!

Only one comment: I would change the comment `# automatic nonuniform FDTD grid with 25 grids per material wavelength` to `[...] per wavelength in the material`.

PR #274, opened 2025-04-09
Legecy Notebook Revamps

tomflexcompute:
We have some notebooks from the super early days and I feel like they are kind of outdated and don't show a good device design. Two particular examples are the ring resonator and linear grating coupler. In this PR I revamped the two notebook so they show the standard design workflow and performance people typically expect. 

Also found a small typo in the y branch notebook that I fixed.


tomflexcompute:
Oh @momchil-flex I used `ModeSimulation` in the grating coupler notebook since we are going to move to that eventually so that we don't have to revise it again in the near future. Should be fine to publish it as like a preview for users?

momchil-flex:
> Oh @momchil-flex I used `ModeSimulation` in the grating coupler notebook since we are going to move to that eventually so that we don't have to revise it again in the near future. Should be fine to publish it as like a preview for users?

It's fine, the only issue is that a user wouldn't be able to see this in GUI yet. So, up to you.

Indeed these notebooks were quite old. I don't even know if we need the uniform teeth grating coupler given that it's not what people usually do, and we have e.g. the focused apodized example. You think it's worth keeping just as a simple starting point? On the other hand, I did like the far fied analysis showing that the emission is around the target angle, wondering why you decided to remove it?

The ring notebook looks great, much simpler and simultaneously useful. The cost has increased quite a lot because it's now 3D but not sure if we care.

tomflexcompute:
> It's fine, the only issue is that a user wouldn't be able to see this in GUI yet. So, up to you.
> 
I'm not too concerned about seeing it in GUI, but I'm more concerned about not having API references available in the docs. Should be fine to just have it in this one notebook. 

> Indeed these notebooks were quite old. I don't even know if we need the uniform teeth grating coupler given that it's not what people usually do, and we have e.g. the focused apodized example. You think it's worth keeping just as a simple starting point? On the other hand, I did like the far fied analysis showing that the emission is around the target angle, wondering why you decided to remove it?

Regarding the far field projection. It might be nice to have but usually if you optimize the coupling efficiency, it simultaneously optimizes the emission angle so people often don't need to check it separately. Since this notebook is a bit long already, I removed it to keep it shorter. I'll add some description in the last paragraph to say that's something doable with our field projection monitor. On the other hand, another good thing to demonstrate, which is also commonly done, is to launch a mode source and take the near field above the grating and then do a mode overlap using the gaussian beam profile. However I think it might be too long to show it. Maybe somewhere else separately. 
> 

In fact I think uniform grating is much more commonly used still since the apodized one can be very sensitive to fab. This design workflow from 2D to 3D is what most people would do when designing GCs. Even for apodized GCs, usually one would do this and then apodize it. 

> The ring notebook looks great, much simpler and simultaneously useful. The cost has increased quite a lot because it's now 3D but not sure if we care.

Yeah I will add a warning in the beginning about the cost. It's not super high and probably as low as we can make it to be in such a simulation.

e-g-melo (approved):
Thanks @tomflexcompute!

I also agree on keeping this simpler grating coupler design. I only have a few comments on it. 
- I recommend breaking the code in cells 9 and 12 into multiple lines.
- Replace `n_c` with `n_sio2` in cell 10.
- " we create a directionary of simulations" ->  "we create a dictionary of simulations"
- The coupling efficiency seems almost constant for source positions between 4 and 5 micron. Maybe a larger span would make it more interesting. 

No comments on the ring example. It is very nice.

alec-flexcompute (approved):
Thanks for doing this revamp @tomflexcompute! In general this is a great revamp. I think the GC example is a good first example of how one can take advantage of our python interface, so I suggest keeping it. I just have a few cosmetic suggestions:

Grating coupler notebook:

1. In cell 4 of the grating coupler, the third line has ## instead of # for a comment. Might be nice-looking to have consistency here, but obviously it super doesn't matter

2. Above cell 9 of the grating coupler, "For the ethed slab mode..." -> "For the etched slab mode..."

3. Above cell 11 of the grating coupler, "making the sweep very efficient" -> "making the sweep very time-efficient"

Microring resonator notebook:

1. Above cell 5, maybe specify that the ring geometry is created from the boolean operation of two cylinder geometries - a new user might think a boolean operation can be done with structures, since earlier in the description Tidy3D Structures are linked.

2. Maybe at the end we can link to an example with our ResonanceFinder object as an alternative way to find Q

Y-splitter notebook:

1. Above cell 5 "using spline for example" -> "using a spline, for example"

2. Above cell 6, center the S bend formula

3. I still think it might be good to include symmetry, because this might be one of the first examples a new user might look at. However, simply linking to our symmetry doc page is fine

tomflexcompute:
Thanks all for taking a look and sharing the feedback.

PR #277, opened 2025-04-14
CI: Add Non-Blocking Notebook Spell Check with PR Commenting

### LINKED ISSUE #207: Notebook linting
Notebook formatting overall is pretty inconsistent, we should probably run `ruff` on the notebooks too.

yaugenst-flex:
Spelling errors within Markdown cells or comments in notebooks currently go mostly unnoticed. We want to catch these errors.

The aim of this PR is to *inform* developers rather than *block* merging, as there can be false positives in the spell check.

**Changes Implemented:**

1.  **`spellcheck.py`:** Executed as part of the `notebooks-linting` workflow.
2.  **Targeted Execution:** The spell check runs *only* on `.ipynb` files that have been modified within the context of the specific pull request. This is achieved using the `tj-actions/changed-files` action.
3.  **Non-Blocking Check:** The spell check step is configured with `continue-on-error: true`. This means that even if `spellcheck.py` finds errors and exits with a non-zero status code, the overall CI workflow will *not* fail. Linting failures will still cause the workflow to fail as before.
4.  **PR Commenting:** If `spellcheck.py` detects any spelling errors or encounters issues processing a file, its formatted output is captured and posted as a comment directly on the PR, see [here](https://github.com/flexcompute/tidy3d-notebooks/pull/277#issuecomment-2808998770) for example.

Closes #207


tomflexcompute:
This is actually great. I ran the spellchecker only on markdown cells instead of the code cells. I'll fix these typos in the other PR.

daquinteroflex:
Went thought the action, nice, quite like the logic! Looks good

daquinteroflex (approved):
Looks good broadly!

yaugenst-flex:
I ran the spell check on all notebooks and fixed the remaining typos. So everything should be clean for now. The remaining spell check errors are false positives that we can ignore. They won't pop up again unless the respective notebooks are changed.

tomflexcompute (approved):
Great addition to the workflow!

PR #279, opened 2025-04-14
add advanced gaussian sources notebook

alec-flexcompute:
First draft - let me know any preferences for plotting or suggestions for a graphic in addition to any other thoughts

### DIFF
--- a/AdvancedGaussianSources.ipynb
+++ b/AdvancedGaussianSources.ipynb
@@ -1,0 +1,1816 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "6b6287fd-21a5-4276-a199-b040bb171541",
+   "metadata": {},
+   "source": [
+    "# Advanced Gaussian sources\n",
+    "\n",
+    "Tidy3D natively offers sources like ModeSource, PlaneWave, GaussianBeam, etc. However, due to the python and open-source nature of Tidy3D, we can easily define arbitrary sources such as vortex beams. In this notebook, we demonstrate how to create these advanced sources in a way that is easily usable in multiple scenarios.\n",
+    "<br><br>\n",
+    "It can often be tedious to create easily usable customized simulation objects from scratch, as there are various details to consider. For instance, if one wants to create an easily-usable beam source from scratch, one must create functionality to automatically interpret the source plane into field coordinates, handle conversion for polar and azimuthal angles, handle different direction specifications, etc. This all becomes very complicated to build, so there is often a tradeoff between ease of construction and ease of use.\n",
+    "<br><br>\n",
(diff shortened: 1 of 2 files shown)


momchil-flex:
Nice! Only comment on my side is to maybe sprinkle some nonzero angles and waist distances in the examples both to illustrate the capability and to test that it actually works as expected. :)

momchil-flex (approved):

tomflexcompute:
Thanks @alec-flexcompute . Very useful tutorial.

1. Add a brief introduction paragraph on these sources in the beginning. Something like "Tidy3D natively offers sources like [ModeSource](), [PlaneWave](), [GaussianBeam](), etc. However, we can easily define arbitrary sources such as vortex beams. In this notebook, we demonstrate how to create these advanced sources for the FDTD simulation..."

2. 'the fact that Tidy3D is open source' -> 'the fact that Tidy3D's [Python frontend](link to github repo) is open source'

3. 'create an Bessel' -> 'create a Bessel'?

4. Not sure if the interactive widget works on the website. Might need to double check. 

5. 'from tidy3d.components.beam import BeamProfile' and 'import ipywidgets as widgets' are repeated in each function. Maybe just move it to the beginning.

6. The usage of 'OAM' appears before `Orbital Angular Momentum (OAM)`.

7. Add the notebook to `docs/features/sources.rst`

8. h2 header capitalization consistency. Either '## Bessel Beam' to '## Bessel beam' or '## Hermite-Gaussian source' to '## Hermite-Gaussian Source' etc.

yuanshen-flexcompute (commented):
Very cool notebook and useful recipes for users working with higher order gaussian beams. My comments are about the layout/presentation:

- Perhaps include more explanation regarding the data structure relationships between the custom BeamProfile object and the method that defines the source and field data. (e.g. what is HGM_obj; what information does it represent/contain; how does one use it after it is defined; what is the difference between HGM_obj and the HGM_source etc.) 
- Similar to Tom's comment, perhaps more elaboration in the preamble to introduce each type of higher order gaussian beam (like a sentence or two to describe what it might be used for; perhaps a figure that shows an example beam profile for each beam type) Wikipedia might be a useful source for this.
- For each example beam, it would be useful to state in text which higher order is being demonstrated (e.g. "this is the l=1, m=2 HG mode")
- Since the notebook workflow is identical for each custom beam type, I think there's no need to repeat the example simulation portion. It is useful to include in the first section (HGM) to show the user how to use the custom beam in their simulation. For subsequent sections, perhaps just include the custom beam method and in lieu of the example simulation, just show a figure of the end product (an example mode). 
- The section titles could use consistent naming, e.g. "Hermite-Gaussian Mode", "Laguerre-Gaussian Mode", "Bessel Mode", "Hyper-geometric Gaussian Mode"

yaugenst-flex:
Could you hold off on merging this until #277 is merged, and then rebase this PR? Would be a nice first test case of the new notebook CI.

alec-flexcompute:
Thanks for the feedback everyone!

@tomflexcompute, for #5, the idea was that a user could just copy/paste the entire cell if they needed a particular source, so I'm hesitant to get rid of the import

@yuanshen-flexcompute for your fourth comment, I still think it'd be good to demonstrate that the field profile propagates correctly for each source, so I kept those example runs.

@yaugenst-flex Lmk when I'm good to rebase/merge!

yaugenst-flex:
Thanks @alec-flexcompute all good! If you could rebase and then wait for CI to run, it'll be ready to merge.

tomflexcompute:
For the OAM beam I expect to see an intensity distribution like this 

image

but it doesn't look very similar in the notebook. Is it because the source plane is a bit too small? 

Also please edit `docs/features/sources.rst` file accordingly.

alec-flexcompute:
@tomflexcompute @yuanshen-flexcompute how does this latest push look?

tomflexcompute (approved):
Great thanks @alec-flexcompute !

PR #281, opened 2025-04-16
Simplify the quick start to the bare minimum

tomflexcompute:
Stripping the quickstart down to the bare minimum as zongfu requested :sweat_smile:


tomflexcompute:
Seems like linting failed but not sure why?

yaugenst-flex:
Actually it's the formatting. Could you run `ruff format` locally? That should fix the issue.

yaugenst-flex (approved):
Oh wow not even a plot with the history :smile:
Yeah this seems to be as basic as it gets

yaugenst-flex:
Actually, if you want to remove one more thing you could do `geometry=td.Box(size=(size_box, size_box, size_box))` instead of `geometry=td.Box(center=(0, 0, 0), size=(size_box, size_box, size_box))`. Center is 0 by default.

PR #282, opened 2025-04-18
Adding the level 2 quickstart notebook

tomflexcompute:
Zongfu suggested a slightly more complex (but not too much) quickstart level 2 notebook so this PR adds an MMI optimization. Most values are again hard-coded to minimize the number of lines...

Also added Yannick's overview article.


Autograd0Overview.ipynb
yaugenst-flex, line 1: I guess this is temporary until we have it on web?

yaugenst-flex (approved):
Nice, I think it looks great!

PR #283, opened 2025-04-23
Added Bragg waveguide filter notebook demonstrating periodic EME

tomflexcompute:


FilipeFcp (approved):
Thanks @tomflexcompute
The notebook is great!

My only suggestion is to add a few more comments on the EMEgrid definition.

I am a bit slow to understand the boundaries and cell definition (so some users might be as well), so maybe just add that the wvg_grid defines the boundaries from 0 to p.

I would also suggest adding a comment on why there is buffering in the x direction, and clarifying that the calculation is only performed between 0 and p.

caseyflex (commented):
Thanks @tomflexcompute , this is a really great example! It is nice to see how you can easily simulate 1000 periods with EME (and even sweep the number of periods). A few suggestions:

- Can you add some discussion about why we use 1 mode and constraint=`unitary`? Without the constraint, power couples into very high-order modes, and this is hard to capture in EME (more modes can reduce the stability of the algorithm, especially if they are not nicely behaved modes).
- I agree with @FilipeFcp 's comment -- in addition, can you maybe say something about the port offsets being zero? I guess this can all be a bit confusing so good to explain it here
- You should probably set `store_port_modes=False` and maybe comment about that reducing the simulation data size and run time if you don't need `smatrix_in_basis` or to inspect the port modes
- I checked the log warnings -- seems likely they will go away after the release, just a version mismatch

tomflexcompute:
Thanks @FilipeFcp and @caseyflex . I revised the EME part a little and added some comments. Let me know if you think the comments are accurate. 

By the way the `EMEPeriodicitySweep` doesn't have an API doc entrance. Also the warning is still there with running from 2.8.3.

FilipeFcp:
Thanks, @tomflexcompute! The new comments are great. If I were using EME for the first time, I would easily understand everything.

caseyflex:
Thanks @tomflexcompute , looks great.

The version warning is because you still have `solver_version=emeperiodic-0.0.0`. But actually I found a minor bug causing it to error under the recently released `2.8.3`. So maybe we just wait until that is patched, then run this without solver_version specified, and then merge the notebook

tomflexcompute:
Yeah it did error without solver specification. It only happens when periodicity sweep is included? Should we push a patch relatively soon?

caseyflex:
I think we will have a hotfix out today

tomflexcompute:
That's great let me know when the hot fix is released and I will rerun the notebook.

caseyflex:
> That's great let me know when the hot fix is released and I will rerun the notebook.

The hot fix is released, try it without `solver_version` and hopefully it works and the warnings are gone.

tomflexcompute:
hmm I ran it without specifying solver version and the warning is still there
image

caseyflex:
Could it be a caching issue? Can you try changing the simulation a little bit?

tomflexcompute:
Ah yes that was why. Now it should be all good. I'll leave it here a bit in case you guys want to take a last look and merge it tonight.

caseyflex (approved):

PR #284, opened 2025-04-24
Add grating coupler optimization with photonforge and autograd example

bzhangflex:


tylerflex:
thanks @bzhangflex ! Looks great. I just had a few pretty minor comments

title : Autograd -> TidyGrad? maybe the way it is now is fine but a couple potential issues:
* will people know what PhotonForge is? not sure if we want to make it more generic. We should at least give a brief intro to what photon forge is in the first paragraph?
* while people know what Autograd / TidyGrad is? maybe we can just say "Inverse design"?
both of these are just discussion points if we want to tweak the title.

[18] (and others) we have an `updated_copy` method that is much easier to use. this would become
```
basesim = basesim.updated_copy(sources=[source])
```
pops up in other places, just FYI

just for aesthetics, it is good to throw a `plt.show()` after plots, otherwise the axes or other matplotlib things like ` <matplotlib.legend.Legend at 0x16aec0740>
` get printed in the notebook.

I think it would be potentially useful to link to the other grating coupler inverse design notebooks. Or even to link to this notebook in the other GC notebooks, just for more visibility. Again an optional suggestion.

Thanks a lot!

bzhangflex:
Thanks @tylerflex! Great points re: the title and introductory information. I tweaked the title and added short descriptions of Photonforge and Tidygrad in the introduction, and also renamed the notebook to "TidyFab0GC" in anticipation of eventually being a part of a TidyFab series. Let me know what you think

tylerflex:
looks great!

tylerflex (approved):

prashkh (commented):
Thanks a lot for this! Looks really great. I have a few comments:
1. Let's add a workflow diagram at the very top of the notebook to explain this relatively long notebook
2. I would give td.Medium(permittivity=1.0, name="air") a name like "air" so later when it's printed it doesn't show all the pole residue stuff
3. You can also print tech.extrusions
4. I would maybe avoid "Currently, adjoint optimization is not integrated into PhotonForge."  We can just say for this workflow we'll use the technology stack to construct a Tidy3D model. The benefit is that you can go from layout to geometry that you can adjoint optimize.
5. For grid_spec there is an easier way to just do by just setting 
pf.config.default_mesh_refinement = 12.0
6. Explain the background simulation a little bit more. So all the things that are not going to change during optimization are built during this step right?
7. Maybe break up the background section to show the background layout component so it's clear you are going from layout (2D) to 3D tidy3d structures
8. Wherever possible link other notebooks or API references
9. For the final result, the improvement with adjoint is not very big. 
I am assuming you already checked different starting points?
How about broadband optimization, can the bandwidth be large while the insertion loss is the same?

bzhangflex:
Thank you so much @prashkh! I think I was able to address all of your comments.

Regarding

> For the final result, the improvement with adjoint is not very big.
I am assuming you already checked different starting points?
How about broadband optimization, can the bandwidth be large while the insertion loss is the same?

Agreed that the improvement is not big. I have tried other starting points and haven't seen better results. I also tried broadband optimization but the bandwidth trades off with insertion loss. I think for the purpose of demonstrating a basic workflow these results are probably ok? I did leave a comment at the end of the notebook challenging the user to try to find a better design though :D.

tylerflex:
Should we go ahead and merge this? @yaugenst-flex

prashkh (approved):
Thanks @bzhangflex for this awesome notebook! It looks great after your latest update. I'll ask Amin if there is a way to link this in our PhotonForge docs as well.

bzhangflex:
Thank you @prashkh! I'll go ahead and merge this then

PR #285, opened 2025-04-24
feature[notebooks]: add robust optimization of directional coupler with sharpness aware minimization optimizer

rahul-flex:
In this notebook, we demonstrate how to design photonic components that remain robust against manufacturing variations—represented by perturbations in the optimizable design parameters—using the Sharpness‑Aware Minimization (SAM) optimizer.

Standard optimization methods often converge to parameter values that yield low loss but reside in "sharp" minima in the loss landscape; such solutions can be highly sensitive to small perturbations and may not perform well when fabricated. Inspired by the connection between the geometry of the loss landscape and generalization, SAM seeks parameter values that lie in "flat" regions, corresponding to neighborhoods with uniformly low loss. Originally developed to improve generalization in machine learning models, SAM achieves this by minimizing both the loss value and the loss sharpness simultaneously. This inherent robustness to perturbations makes SAM suited for designing photonic devices that are tolerant to manufacturing errors, such as geometry variations in lithography.   

We will apply SAM to design a robust 50/50 directional coupler (DC), optimizing the boundaries of the coupling region defined using PolySlab structures within Tidy3D. This approach aims to produce a DC design whose performance is stable across a range of likely manufacturing variations in coupling region boundaries, drawing on the principles outlined by Foret, P., Kleiner, A., Mobahi, H. and Neyshabur, B., 2020. Sharpness-aware minimization for efficiently improving generalization. arXiv preprint arXiv:2010.01412.

Results:
We optimize for the split ratio using SAM and analyze the robustness of initial design and optimized design for geometric perturbations:
image


tylerflex:
what's the status of this notebook? should we merge or does it need more work?

yaugenst-flex:
@tylerflex unsure, the main problem is that i think the results are hard to reason about and they don't seem very convincing. maybe you have some thoughts? it is a demo for a different optimizer, but not sure if it makes for a great example..

tylerflex:
yea I see what you mean. I wonder if there's some way to still publish this somewhere? even if not directly in our docs?

tylerflex:
@yaugenst-flex I suggest we close this issue for now. we can re-open if we decide to purse cleaning it up? feel free to to reopen if you disagree (doing some spring cleaning)

PR #287, opened 2025-05-01
adding alternative mode volume calculation for lower memory usage

FilipeFcp:
Hi all.

Since a user used this notebook as a reference and had issues with kernel crashing during mode volume calculation, I made a small modification to reduce the memory usage, and also added a small discussion on how to use raw monitor data (without expanding symmetries) to calculate it.


yaugenst-flex (commented):
Thanks @FilipeFcp looks good to me overall. Only thing I'm wondering is whether the `interval_space` change was necessary? It seems like you changed from coarse spatial sampling to coarse time-domain sampling, with a net memory decrease of a factor of 2, which sounds reasonable but I'm wondering keeping the coarser spatial sampling would change the results significantly?

tomflexcompute (commented):
That's great @FilipeFcp . Seems like there is a 10-20% difference in the calculated mode volume. Are there results in the literature where we can benchmark the results?

FilipeFcp:
Thanks @yaugenst-flex 

It doesnt change significantly, but it brings the results closer to those reported in the paper. For this type of calculation, I believe the error introduced by using the raw data is preferable to the loss of data caused by downsampling.

I added a comment regarding the interval_space in the notebook.

FilipeFcp:
@tomflexcompute Yes, the structure is from [this paper](https://doi.org/10.1063/1.3107263). They report 0.39 (lda/n)^3

tomflexcompute:
> @tomflexcompute Yes, the structure is from [this paper](https://doi.org/10.1063/1.3107263). They report 0.39 (lda/n)^3

Great that's very good!

tomflexcompute (approved):

yaugenst-flex:
@FilipeFcp if you could squash these commits it’s good to go

PR #288, opened 2025-05-05
Quick example with caveats for simulating low-index contrast waveguides

FilipeFcp:
Hi everyone,
We had two support cases last week with a similar issue, so I thought it would be helpful to create a quick guide. I’m not sure if we publish it like this in the learning center or if we should just add it to an existing resource.

Let me know what you think!


tomflexcompute (commented):
Thanks @FilipeFcp . A very common issue from users and this can serve as a great tutorial.

1. h1 header should be placed above the library import cell.Also maybe make the title a bit more informative like "Avoiding unphysical loss in low contrast waveguides"? 
2. "with more than three times less Flexcredits consumption." might be better rephrased as "while consuming less than one-third of the Flexcredits." or "with Flexcredits consumption reduced by more than threefold."
3. Do you have any preference on where to display this notebook? Maybe under Advanced Applications? Could you edit the rst file accordingly?
4. In the introduction, add a sentence about why low index contrast waveguides require special attention in this regard, i.e. in low contrast waveguides, especially when the waveguide is narrow like the tip of the edge coupler, the field can be very unlocalized with a large mode field size.
5. Add some schematics or plots to help visualize the simulation.

alec-flexcompute (approved):
Excellent notebook as always @FilipeFcp ! A couple things:

1. My rendering of the notebook has some weird line breaks at the beginning/end of each cell. Might want to check if this is the case for you, and if so, remove them
image
2. Add a plt.show() at the end of the cells for cells 5 and 10
3. Mild suggestion that can be discarded: maybe plot an example simulation created from the "get_sim" function so it's easier for the audience to visualize

Other than that, clear, concise, very nice

FilipeFcp:
Hi all. I think I addressed all.

@tomflexcompute, I just added a simple plot. Do you think a more detailed schematic would be better?
@alec-flexcompute, this is odd, but I am seeing the same thing in all notebooks:

image

tomflexcompute:
Thanks @FilipeFcp looks very good. It might be worth it to show the mode profile to visually explain why a larger domain size is needed. Also use the mode profile to determine what the appropriate size is. 

I would suggest again making the title a bit more informative like "Avoiding unphysical loss in low contrast waveguides".

FilipeFcp:
Thanks @tomflexcompute. Show the modes is a good idea 
Sorry, I somehow skipped the title part in my first review.

tomflexcompute (approved):

PR #289, opened 2025-05-08
Tom/disordered plasmonics

tomflexcompute:
Added a new example notebook to address the question of chromaticity calculation and diagram plotting. 

As discussed in the channel @momchil-flex the result is quite sensitive on the grid size.


FilipeFcp (approved):
Great notebook as always @tomflexcompute !

Just a few comments:

Before cell[2]: "we will simulation" → "we will simulate"

In cell[5]: "discorder" → "disorder"

Just out of curiosity, why not use an override region?

Totally optional: sometimes I like to use the `shading='gouraud'` argument with `pcolormesh`. I think it looks a bit prettier.

tomflexcompute:
Thanks @FilipeFcp . All fixed. 

Using mesh override on the nanoparticle region doesn't help too much with convergence I found. Nonetheless I agree it might be good to show it anyway so I took this approach in this revision.

e-g-melo (approved):
Hi @tomflexcompute! Thanks for this amazing notebook. It will be pretty helpful for those working on structural colors.

I don't have any additional comments about the work. 

Totally optional: it would be interesting to see some reflectance spectra and their respective RGB colors. :)

tomflexcompute:
Thanks @e-g-melo . That's a great idea. I've added the actual color plot in the end.

momchil-flex (approved):

PR #290, opened 2025-05-09
Charge solver update:

marc-flex:
- Added refinement regions
- Charge simulation is now 2D
- It uses the perturbation from the convenience function
- Updated the definition of `w_core` to match the figure

@momchil-flex @alec-flexcompute Since there isn't currently a released version that incorporates the mesh refinement functions I'm keeping this as Draft, but do feel free to go through it with comments


alec-flexcompute (approved):
Looks good @marc! One question - when you say that the charge simulation is now 2D, what do you mean? It was 2D in this example before

marc-flex:
@alec-flexcompute I just realized I didn't respond to your comment. Before it had a very small size in the z direction but it was 3D. Now it is fully 2D.

alec-flexcompute:
Gotcha, thanks!

momchil-flex (approved):
Great, seems to me it can be merged after squashing!

PR #293, opened 2025-05-19
refactor: remove code referencing gdspy due to its removal from tidy3d

frederikschubertflex:
This is the accompanying notebook change for the PR https://github.com/flexcompute/tidy3d/pull/2477.


.binder/requirements.txt
daquinteroflex, line 1: I don't think we actually have binder working but it'd be welcome if we can fix that.

daquinteroflex (changes requested):
Could we add a warning at the top of the notebook saying as of x version we've depreciated gdspy?

.binder/requirements.txt
frederikschubertflex: Good idea! I created https://github.com/flexcompute/tidy3d-notebooks/issues/294 for tracking this.

frederikschubertflex:
I added a warning to the notebook introduction.

daquinteroflex (approved):
Great thanks!

PR #295, opened 2025-05-22
Added the TPA/FCA notebook

tomflexcompute:
Applied some polishing and cosmetic changes to @momchil-flex 's version. Amazing results!


caseyflex (changes requested):
This is great! Thanks @momchil-flex and @tomflexcompute for your efforts.

- Might say something about where the 8/3 and 4/3 are coming from in the implementation equations (https://en.wikipedia.org/wiki/Kerr_effect#AC_Kerr_effect), and maybe reference the paper https://ieeexplore.ieee.org/document/4299014
- Scaling sigma and tau inversely is a clever way to reach the same steady-state behavior faster. You can see this very nicely from the equation for steady-state Nf. Maybe we could make a similarly clear argument for scaling beta and the waveguide length inversely. There is a nice formula here: https://en.wikipedia.org/wiki/Two-photon_absorption#Absorption_rate. 
image
comes from solving the diffeq dI/dz = -beta I^2
- Can we add an AuxFieldTimeMonitor to record and plot Nf, as an excuse to demonstrate this functionality? And because it is nice to see Nf too.
- I still don't like turning off warnings. Maybe we can leave the warnings for now, and let's separately consider turning off field decay warnings when there are CW sources. Also it says "for the simulations below" but there aren't any?
- The paper specifies a linear optical loss, should we incorporate this into the material model? We should also scale it, as we scaled beta.

This looks really great! Finally we can have a good notebook for this feature...

momchil-flex:
Thanks @caseyflex , great suggestions!

> * I still don't like turning off warnings. Maybe we can leave the warnings for now, and let's separately consider turning off field decay warnings when there are CW sources. Also it says "for the simulations below" but there aren't any?

The warning appears for every simulation data from the batch when the results are loaded.

> * The paper specifies a linear optical loss, should we incorporate this into the material model? We should also scale it, as we scaled beta.

Yeah I guess we could, it's just pretty small so I don't think it will affect things much, but for completeness I guess we should.

caseyflex:
For FCA in steady state, there is a Beer’s law dependence e^(-alpha x) with alpha proportional to sigma tau beta E^4. So this is also handled correctly with the length and beta rescaling.

well I guess the full dependence not assuming small total loss from FCA is a nonlinear diffeq, but I think the scaling argument still holds

caseyflex:
One more suggestion is that maybe instead of plotting the instantaneous flux, you could plot the cycle-averaged flux, or the peak power within each optical period.

caseyflex (approved):

caseyflex (approved):
Looks good. One final comment is you might mention why the aux field is "Nfy", namely that the different components don't interact nonlinearly as a modeling approximation.

tomflexcompute:
@momchil-flex any additional comments?

momchil-flex:
Looks good! Maybe just one extra sentence about scaling `beta`, related to @caseyflex 's comment above. Maybe no need to go in too much detail but just: 

```
For this illustration, we also will not be simulating the whole 4.8cm waveguide from the paper. We can scale the TPA coefficient , making the overall nonlinearity stronger, while correspondingly shortening the waveguide. This is a valid modification to the setup since both the two-photon absorption, and the free-carrier concentration Nf, and thus the free-carrier absorption, are proportional to beta.
```

PR #296, opened 2025-05-26
Updated TPA notebook header to h1 level

tomflexcompute:
This PR fixes the TPA notebook title, which needs to be a h1 header to be rendered correctly on the docs and learning center.


e-g-melo (approved):
Very nice notebook!

PR #297, opened 2025-05-27
add dispersion model for comparison and use a few new API improvements

### LINKED ISSUE #245: Update impedance notebook to apply a frequency-dependent model
Proposed by Yangchen: to use a frequency-dependent model (such as Fig. 9 in [here](https://ntrs.nasa.gov/api/citations/19900004378/downloads/19900004378.pdf)) as instead as groundtruth

dmarek-flex:
Big change is adding the dispersion model from scikit-rf. I also changed to a 3D PEC for the strip, since our conformal meshing is much improved.

Also added section describing the different definitions of impedance commonly used.

- [x] Need to rerun with the next version of the solver (2.8.4 has some bugs which cause the notebook to error)


daquinteroflex:
Looks good! Would it be possible to use the `skrf` prefix instead of `rf` because that way it's easier to work out what's external to tidy3d since `rf` and our `microwave` would appear to be from the same package.

daquinteroflex (approved):

weiliangjin2021 (approved):
Looks great!

- "Internally, the ImpedanceCalculator computes impedance using ..." maybe change it to mention that it is one of the 3 approaches
- For Gaussian pulse, how is the result if it's constructed from `from_frequency`?
- Could you comment how the 3 methods compared to skrf result?

dmarek-flex:
> For Gaussian pulse, how is the result if it's constructed from from_frequency?

Basically no change

dmarek-flex:
> Could you comment how the 3 methods compared to skrf result?

Done, I think I found a better way to compare the methods and explain the differences when compared to scikit-rf. I want to still finalize the last cells to show a calculation of a lossy transmission line.

weiliangjin2021:
> > For Gaussian pulse, how is the result if it's constructed from from_frequency?
> 
> Basically no change

In that case, better to use `from_frequency` to promote this approach?

dmarek-flex:
> > > For Gaussian pulse, how is the result if it's constructed from from_frequency?
> > 
> > 
> > Basically no change
> 
> In that case, better to use `from_frequency` to promote this approach?

Ah, ok! It did change the `FieldTimeMonitorData`, so I'll need to edit those plots.

tylerflex:
should we merge this?

dmarek-flex:
> should we merge this?

I was waiting for the next version of the solver to rerun, so I can remove the temporary `solver_version`. I think I can do that with the current pre release.

PR #298, opened 2025-05-30
Added topological photonic crystal notebook

tomflexcompute:
Added topological photonic crystal notebook.


FilipeFcp (approved):
Thanks @tomflexcompute 
This paper is pretty cool!

My only comment is that there is an empty cell at the end.
Everything looks great and clear.

tomflexcompute:
Thanks @FilipeFcp . I think `ruff` intentionally creates an empty cell at the end.

momchil-flex:
Thanks for your help @tomflexcompute ! I like your decision to plot the flux ratio in the end!

Just one comment on my side: I wanted to say that the return of this function

```
def make_unit_cell(s: float, a0: float, r: float, h: float, x0: float, y0: float) -> "td.PolySlab":
```

should be a list of PolySlab rather than a single one. But then I was a bit confused because you start with `triangles = 0` and then you `+=` the PolySlab-s, so I guess the return is a "union" ClipOperation of polyslabs? Overall I was actually surprised that you can start with `0` and then add geometries, maybe this is a known thing we do in other notebooks too though. But I wonder if performance and clarity wise we shouldn't prefer either just list of geometries, or a GeometryGroup for such simple unions.

tomflexcompute:
> Thanks for your help @tomflexcompute ! I like your decision to plot the flux ratio in the end!
> 
> Just one comment on my side: I wanted to say that the return of this function
> 
> ```
> def make_unit_cell(s: float, a0: float, r: float, h: float, x0: float, y0: float) -> "td.PolySlab":
> ```
> 
> should be a list of PolySlab rather than a single one. But then I was a bit confused because you start with `triangles = 0` and then you `+=` the PolySlab-s, so I guess the return is a "union" ClipOperation of polyslabs? Overall I was actually surprised that you can start with `0` and then add geometries, maybe this is a known thing we do in other notebooks too though. But I wonder if performance and clarity wise we shouldn't prefer either just list of geometries, or a GeometryGroup for such simple unions.

Actually the returned `triangles` from all the clip operations is a `GeometryGroup` object already, the type hint was indeed incorrect. I updated it and also made the geometry group definition clearer by putting all the triangles in a list and then creating a geometry group explicitly.

tomflexcompute:
Should be good to go?

momchil-flex (approved):
Yep!

PR #299, opened 2025-06-04
uploading directional scattering notebook

FilipeFcp:
Hi all,
I’ve reproduced the main steps from [this](https://pubs.acs.org/doi/10.1021/nn402736f) paper. The main idea is that by tuning the magnetic and electric dipole components of the scattering cross section, it’s possible to suppress backscattering reflectance. The paper is a bit confusing in that it performs the mode decomposition for a single sphere, which doesn’t directly apply to an array. This makes the results feel somewhat disconnected. I only noticed this while revising the notebook. Nevertheless, the results match the paper quite well, so it is a good comparison against CST. 

I also adapted the multipole expansion notebook to account for a background medium different from air.


alec-flexcompute (commented):
Great job as always @FilipeFcp ! A couple suggestions:

1) (very mild suggestion) maybe specify that you're varying the aspect ratio by varying the radius at the start

2. label the axes on the last two plots

Cool visualization on the last plot!

tomflexcompute (commented):
Thanks @FilipeFcp very interesting model and a great use case of multipole expansion. 

1. Maybe change the title to "Tailoring directional scattering in silicon nanodisks"?
2. "function for carry out multipole decomposition" -> "carrying out"
3. "results close match" -> "results closely match"
4.  Not sure if it's just the github web viewing issue but there is quite a lot of empty space above and below the thumbnail image. 
5. I didn't check but I assume the metadata have been added properly.

e-g-melo (approved):
Thanks @FilipeFcp! This is quite an interesting example.

1. Starting with cells 4, 5, and 6 under a "Simulation Setup" H2 header would be beneficial. This approach can help users grasp the context before navigating through the dipole expansion functions.
2. An H2 header before cell 7, along with an explanation of the purpose of the parameter sweep, would also help users follow the notebook.
3. I would recommend moving the comments to the beginning of the cells in markdown format. For example, the comments on cells 7, 8, 9, and 10.
4. There are missing labels in the chart of cells 12 and 16.

FilipeFcp:
Thank you @alec-flexcompute @e-g-melo and @tomflexcompute for the comments.
I believe I have addressed all the points.

tomflexcompute (approved):
Thanks @FilipeFcp . Looks good to go.

PR #305, opened 2025-06-10
Add absorption power computation to the Plasmonic Nanoparticle notebook

damiano-flex:
This PR introduces the computation of the total power absorption and the absorption power density in the Plasmonic nanoparticle notebook. The changes are the following:
- Added a small markdown text explaining how the absorption power is computed.
- Added a snippet where it is computed the power density, and then the total absorption power through an integration.
- Added the plot of the total absorption power, which matches perfectly with the other computed one from Tidy3D.


momchil-flex:
Thanks @damiano-flex this is really great! Here's a few comments:

- In the `Create monitors` cell, you can add an item about why we are adding 3D field monitors, namely saying that we will use those to compute the absorption distribution within the particle, and the total absorption by integrating that.
- I would like this tutorial to also serve as a tutorial on how to compute D, as well as quantities that depend on D. So I suggest a) use the E.dot(D) formula rather than |E|**2; b) add an explanation that to compute D accurately, we use `colocate=False` in the E-field monitors, and combine the data with the permittivity monitor data
- Maybe instead of `Compute the power absorption`, `Absorbed power distribution` or something like that? Because in the cell right above we already get the total `absorbed` power from the flux monitor.
- Let's move the field plots to the `Absorbed power distribution` section and add a new plot with the absorption distribution after it is computed.
- Final section name can then just be "Comparison with Mie series" as there will just be the Mie plots left there.

damiano-flex:
@momchil-flex I addressed your comments. Let me know if this version is OK.

momchil-flex (commented):
Thanks for the revision! Almost there!

- In the permittivity monitor discussion, "`colocate` is set to `False`, since the permittivity is inherently defined at the center of each grid cell." -> "`colocate` is set to `False` so that the permittivity at the Yee grid locations, exactly as used in the solver, is stored.
- Could you redo the `web.run` call so the outptut is actually shown?
- In the absorption spectrum labels, how about we use "Tidy3D: $P_{abs}$" and "Tidy3D: I - S", and comment in the text what `I - S` (incoming minus scattered power) stands for.
- Can we give a reference for the absorbed power density?
- I know this didn't come from you, but in the end, in this sentence "Since the sphere's material is dispersive, no subpixel averaging scheme is applied", let's replace "dispersive" with "metallic". We actually apply subpixel to dispersive *dielectrics*.

momchil-flex (approved):
I'm clicking "approve" but I have one final comment. I don't think deriving the E*.dot(D) formula from Im[eps] |E|^2 is conceptually right - the former is more fundamental than the latter. From Gemini:

image

While your discussion is not technically wrong, I would still flip it around. In principle it's not even needed to mention the |E|^2 formula at all, but it could serve a good purpose to explain that in that way it is easy to see that absorption can only happen where Im[eps] != 0.

PR #307, opened 2025-06-11
adds adjoint to metalens notebook

tylerflex:
This adds a new section to the visible light metalens where we use local field projection and adjoint to fine tune the power directed to the focal spot.

We get a 17% improvement over 15 iterations.

NOTES:

* uses `yaugenst-flex/polyslab-grads` frontend branch, which has not been merged. without the code from this branch, the results will be terrible because the gradients are inaccurate.
* uses local gradient calculation to make use of the changes in this branch. Ideally we would switch to remote gradient.
* The gradient processing takes >5 minutes each iteration.. so I limited the code to only run one iteration. I left the cell output to correspond to the full 15 iterations that I ran locally, along with a small note.

Not sure when is the best time to merge this, maybe after Yannick's PR is merged and it is fast enough to justify running this?


tylerflex:
@tomflexcompute this should be ready to go, I fixed the linter error and made sure it ran properly on the latest develop branch.

tomflexcompute (approved):
Thanks @tylerflex good to go after you correct the typos AI suggested.

tylerflex:
Thanks @tomflexcompute , fixed all relevant typos. Apparently I have a really hard type spelling the word "function", seems like every notebook I spell that wrong.

I will merge

PR #309, opened 2025-06-17
Add custom heat source notebook

damiano-flex:
This PR implements the example described in [996](https://github.com/flexcompute/tidy3d-core/issues/966), by adding a new notebook that demonstrates how to set up and run a simulation using a custom (point-wise) heat source.

The main objective is to show users how to import spatially-varying power data from an electromagnetic simulation to define a pointwise heat source for a subsequent thermal analysis.

Specifically, this example:

- Reproduces the photothermal heating results of a gold nanosphere from the paper: [Multiphysics Modeling of Plasmonic Photothermal Heating Effects in Gold Nanoparticles and Nanoparticle Arrays](https://doi.org/10.1021/acs.jpcc.0c02443).
- Shows the complete workflow, starting with an electromagnetic simulation in Tidy3D to compute the absorbed power density of the nanoparticle.
- Demonstrates how to use the output of the electromagnetic simulation as an input for the heat solver to accurately model the temperature increase.


momchil-flex (commented):
Thanks @damiano-flex ! Overall this is really great!

A few comments that I noticed right off the bat:

- The top level title of the notebook should not be `Introduction`, but an actual title of whatever is going to be studied. This is what will appear in our documentation table of contents, not the file name of the notebook.
- Instead of throwing random 1e-3 factors around, you should be able to just scale the source by adding an `amplitude` in the time dependence: `gaussian = td.GaussianPulse(freq0=f0, fwidth=fwidth, amplitude=np.sqrt(1e-3))`
- Not sure if you use cells in sphere of 80 because I changed that at at some point but the results might be pretty similar at 50 already
- For the absorbed power computation, you should add a little bit of text, but mostly we should reference the other notebook #305 , once it's published. Could you also try to avoid the code duplication for the absorption computation and instead provide a function for it that is called for each particle?
- I would suggest not doing horizontal figsize larger than 12, or maybe 14 tops. `figsize=(18, 7))` which you currently have would still squeeze the figure to the notebook size, but make the font size of the figure annotations small.
- "However, since the current solver only calculates the final steady-state temperature distribution, the capacity parameter is ignored. We've included it here for completeness and compatibility with future transient solvers." -> we've already released a transient heat solver.
- We are deprecating using `Medium.heat_spec` in favor of using heat-specific [AbstractHeadMedium](https://github.com/flexcompute/tidy3d/blob/develop/tidy3d/components/material/tcad/heat.py) subtypes. Even more generally, one could define from the start a `MultiPhysicsMedium` with `optical` and `heat` properties that will be used in each of the solvers. You could then also define the scene from the start. But this is not necessary to do if you think it's better to first talk about optical only and only then introduce heat.
- You should mention somewhere that the results agree with the paper very well :) but I think we generally don't copy over figures from there, so just in words.

I'm adding @alec-flexcompute and @tomflexcompute here since they know better our review rules for notebooks, and also since they'll be developing more heat solver examples.

tomflexcompute:
Thanks @damiano-flex and congrats on the first notebook PR! Great feature to showcase and you definitely picked a great paper for reference. Some comments, mostly regarding formatting for you to consider:

- For the notebook title we only sentence capitalization (only capitalize the first character of the first word). For section headers (h2 or higher), we actually use title format (capitalize the first character of each word). See other notebooks in the repo for reference. This is a bit of a confusing format but somehow we have followed this since the beginning. 
- In the second cell the comment says "radius and location of the nanoparticles" but it only defines the center without the radius. Radius is defined in the next cell instead.
- In the markdown above permittivity and field monitor definitions, comment on why we need to set colocate to false.
- In the markdown you mentioned using a "adaptive meshing", which usually means something a little different. To avoid confusion, I would suggest just say something like "nonuniform mesh".
- Since the simulation with the rod is named sim_rod, source_rod, etc, it might be more clear to name the one with sphere something like sim_sphere, source_sphere, etc?
- We need to add metadata to the notebook for SEO and categorization purposes. See this [guide](https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e) for instructions. 
- Also we need to add the notebook file name to one of the rst files. This notebook can be in `docs/case_studies/nanophotonics.rst` or `docs/features/heat.rst` depending on if you want to categorize it as a tutorial or a case study. Either is fine in this case I think.
- Last point for @momchil-flex is the power absorption density calculation is still a bit tedious. It would be great to have a dedicated monitor or postprocessing convenient function to streamline this for users.

alec-flexcompute (commented):
This is a great notebook @damiano-flex !

In addition to Momchil's feedback, I have a few suggestions, mostly cosmetic:

1. When defining the permittivity monitor, define the center first and then the size to match the definition of the Field Monitor

2. It might be useful to describe buffer_override in the meshing stage. Just a 1-sentence comment above would do

3. Remove the comment about the uniform grid in the first simulation definition

4. When computing the volumetric power absorption density, I don't think the comment or the use of the "nearest" method should be included. Also, in general for examples we should use plural first person instead ("We" vs "I")

5. Under the "Heat Simulation" title, I'm not sure what you mean by "We will now define the physical parameters for our simulation". Also "the key material property" -> "the key material properties"

6. In the cell where you define the material heat properties, "1. Define Materials" -> "Add material thermal properties". Also the comments about the heat capacity not having effects in the version state of the solver, could you clarify this? It seems to contradict the description above. (also there's two dashes in the second comment)

7. "It is strongest in regions where the electric field is most intense, and weaker in regions where the electric field is least intense." -> "It is strongest in regions where the electric field is most intense, and weakest in the regions where the electric field is least intense."

8. Maybe instead of source names "source1" and "source1_custom_data" you could have "source_sphere" and "source_sphere_custom_data"

9. Also label the heat simulations with more specific names, eg "heat_uniform_sphere" and "heat_point_sphere"

10. For the final conclusion, maybe add "when irregular particle geometries are used"

11. Mild consideration: Instead of creating heat simulations from scratch, you might want to update the EM simulations with structures built by using updated_copy to add the heat_spec mediums to the EM simulation structures. It might be cleaner this way, and suggest higher coupling between EM and Heat.

Once again, great job!

marc-flex (commented):
This is really good @damiano-flex!

Apart from the comments below, I'd say that we should warn the users this may take a few hours to run and about 5 FC. In fact, I'd be tempted to decrease the cost of simulations themselves and at the end say something along the lines of "this has used low resolution simulations for convenience. If we increase the resolution ...the following results are obtained which agree with the reference paper...". Although maybe just warning users may be enough. 

- `...we consider a Golden Nanosphere (GNP) and with a radius of 68 nm and Golden Nanorod with a radius of 26, 91nm height and 11nm of the radius tip. ...` -> `...we consider a Golden Nanosphere (GNP) with a radius of 68 nm and Golden Nanorod with a radius of 26, 91nm height and 11nm of the radius tip. ...`
-  `Tidy3D's built-in plotting methods provide a quick and easy way to inspect the simulation layout in 2D or 3D.` here you mention we can inspect the simulation layout in 2 and 3D but then you only show 2D views. Maybe add a comment saying something along the lines of `Given the simple geometries of the example we only show 2D plots. For more information about 3D see....`
- The grid seems a lot finer in the rod case. Why is that? Is it because of the accuracy you mentioned our private conversation?
- I think we should reuse the structures. This is, you could add `heat_spec` to the mediums at the top, so we don't need to replicate?
- In the sentence `allowing us to compare a simplified approximation with a more physically rigorous approach` I'd use "compare against" rather "compare with" but maybe ask some English speaking person better. @alec-flexcompute ?
- In `Uniform Heat Source : Simpler and less accurate`. Uniform doesn't automatically mean less accurate. It's closer or further from the actual physics. In fact, by looking at the results it seems like a perfectly good approximation.
- In `Point-wise Heat Source : More sophisticated and physically accurate.`, I'd say something like `Spatially distributed heat source: more complex and closer to the problem physics.`. In general I'd avoid using the phrase "Point-wise" since it sounds like you have a collection of point sources which isn't technically true.
- Since the plots with constant vs. dataarray look the same, maybe a more interesting figure would be the difference between the two? We could inspect the actual effect and quantify how/where they differ. So I'd have: the temperature field for the sphere and a plot with deltas; and the field for the rod and its deltas

alec-flexcompute:
@marc-flex  "compare against" is around the same as "compare with" - I don't think it matters

damiano-flex:
Hello everyone! thank you for the feedback. I addressed all the comments. I also renamed the file to match with the final title. Let me know if this version is PR-ready

alec-flexcompute:
Looks good, except I think the first edit I suggested still stands

marc-flex:
Thanks @damiano-flex! This is looking very good!

I guess one last comment: you still have this in your temperature plots: `Temperarature plot with pointwise absorbed power density source`. Maybe just change these titles to something like

- `Temperature distribution using uniform heat source`
- `Temperature distribution using spatially distributed heat source`

These are maybe more in line with the current text?

tomflexcompute:
Thanks @damiano-flex looks very good. If we put this notebook in the case study category, it does require a thumbnail image. I created one version here but if you would like to explore other options feel free to do so. 
particle_heating

The thumbnail image can be added in the begining of the notebook and more importantly in the metadata so the [learning center](https://www.flexcompute.com/tidy3d/learning-center/example-library/) can be built correctly.

momchil-flex (approved):
Approving assuming you'll address the last small comments, thanks!

damiano-flex:
Removed pointwise everywhere, and added the picture + associated metadata.

tomflexcompute (approved):
Thanks @damiano-flex !

PR #310, opened 2025-06-19
updated notebook

FilipeFcp:
Hi @marc-flex ,

I’ve updated the notebook. In cell [16], I now assign the absolute value of the current, so everything works correctly. That said, it’s still a bit confusing to me why the current is negative and doesn’t change when I invert the bias.

I also made a few minor updates, removed unused comments,  changed deprecated classes, and updated the charge simulation to use the `.from_scene`, for consistency with the rest of the notebook.

Additionally, I’m uploading another version (not meant to be merged) that uses the new spatially varying heat source. I thought it could be a good opportunity to showcase that feature, but the results seem a bit off. I also feel there might be a better way to implement the calculation.

On a related note, I think it would be great to have a version of the notebook where the Heat and Charge simulations are run together. That could be a compelling example for customers. Do you think that's feasible in this case? I gave it a try, but it only seemed to work under specific mesh conditions, and I couldn’t find the Temperature monitor in the simulation results.

Sorry for the big text, and thanks in advance!


marc-flex:
Thanks @FilipeFcp for having a look at this!

So the ThermoOpticDopedModulator is a p-n-p type device which means it really doesn't matter if you apply the bias to one side or the other, you'll get the same result. That's why you always get negative voltage, it's symmetric. In other words, if you apply the voltage to the right and the current flows left (you see negative because you're obtaining current on the right) when you apply it to the left, the current flows right (and since now you're obtaining current on the left, again you see negative). 

In that sense, it makes sense to have the absolute value, as you have done. 

Apart from that maybe a couple comments on the ThermoOpticDopedModulator notebook:

- There are couple of cells 12 and 13 that seem to be for your own testing (they're just a couple of tuples)
- You need to redefine the simulation center when creating the charge simulation. You're cutting the waveguide. I have tried with `center=(0, h_core/2, 0)` and that seems to be OK.

marc-flex:
Thank you also for creating the distributed heat source! It's very cool. However, given the assumptions made, I think it's bast if we hold back with that one. We're working on a non isothermal solver which will provide also self heating temperature.

FilipeFcp:
Hi @marc-flex 

Thank you for the comments.

I think it is worth adding a comment regarding the current. Just to make sure I understand: Is the sign convention that current flows toward the higher potential? If the carriers were electrons, would the current always be positive? Also, I tried a negative bias, and the simulation errored.

Regarding the .from_scene methods, do they only import the structures?

And another question: the warning
WARNING: Structure at 'structures[0]' has bounds that extend exactly to simulation edges.
Is this just inherited from the optical solver, or does it have some meaning in charge and heat simulations?

Sorry again for the many questions, but since I am working on the FAQ, it is a good opportunity to understand everything better.

marc-flex:
I tried with negative bias and it worked. What are you trying?

In terms of sign, with current sign convention electrons flow towards lower potentials.

If I'm not mistaken, it imports both the structures and the background medium.

Yes, that warning is inherited. 

Do ask as many questions as you need! :smile:

FilipeFcp:
Just this: voltages = `list(-np.arange(0.0, 5.0, 0.5))`
[This](https://admin.simulation.cloud/monitoring/tidy3d/tasks/user-c8a8b545-588d-40dc-953c-31fcb17c93f4/hec-65e8ac2d-9db0-4887-8084-1135ba9b63f6) is the task

marc-flex:
> Just this: voltages = list(-np.arange(0.0, 5.0, 0.5))
[This](https://admin.simulation.cloud/monitoring/tidy3d/tasks/user-c8a8b545-588d-40dc-953c-31fcb17c93f4/hec-65e8ac2d-9db0-4887-8084-1135ba9b63f6) is the task

Right. The problem is the -0 it is creating. You can instead try `list(np.arange(0.0, -5.0, -0.5)[::-1])`

FilipeFcp:
Thanks @marc-flex 

I initially thought the `.from_scene` would also define the simulation size and center, but I’ve now set the correct center manually.

I also removed the extra cells and added a brief comment regarding the warning.

On a side note: do you have any plans to include a built-in interpolation option in the updated_copy method in the future?

marc-flex:
> On a side note: do you have any plans to include a built-in interpolation option in the updated_copy method in the future?

What do you need to interpolate?

marc-flex (approved):

PR #316, opened 2025-06-30
:books: Update headers to use-cases

daquinteroflex:
Not updating toctree folder path because that involves redirects and not super sure it is required.

### DIFF
--- a/docs/case_studies/index.rst
+++ b/docs/case_studies/index.rst
@@ -1,5 +1,5 @@
-Case Studies
-------------
+Use Cases
+---------
 
 .. toctree::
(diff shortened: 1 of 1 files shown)


tomflexcompute (commented):
Thanks @daquinteroflex I remember at some point we needed to make the dashes the same length as the text. Is it still the case?

tomflexcompute (approved):
@daquinteroflex I will approve the PR but feel free to make sure the dash length is not an issue.

daquinteroflex:
Updated it! Will merge for next release

PR #317, opened 2025-06-30
Update AntennaCharacteristics notebook

yuanshen-flexcompute:
I have updated this tutorial to better tutorialize our new RF and antenna modeling functionality. I have also incorporated comments from @athielens-fc (thanks, Arno). 

Compared to the old tutorial, I have made the following key changes:
* Added metal thickness to showcase `LayerRefinementSpec`
* Modified frequency range to focus on first two resonances. Reason being that this is tutorial rather than a benchmark, so we should focus on a simple user scenario rather than the exact paper parameters.
* Added far-field metrics section in Results to explain how to access all commonly used antenna metrics
* Added impedance calculation
* Modified some far-field plots. 
* Removed explanation and equations of the far-field metrics. Reason is that the target audience is someone who is newer to Tidy3D, but not new to antenna theory. 
* Removed left/right polarized far-field section. I didn't feel like these plots were showing anything meaning full. Please correct me if I am wrong. 

Please let me know if you have any suggestions to further improve this tutorial for the beginner antenna simulation user. In particular, I am looking for feedback in the results section for more interesting far-field data to present.


alec-flexcompute (approved):
Nice job @Yuanshen on this revamp. A couple suggestions:

1. A bit more explanation on what de-embedding is doing would be helpful (I'm still learning microwave electronics myself though, so I'm not sure how necessary this is)

2. I think it'd be useful to plot one or two of the far field metrics just to demonstrate how to do so - the antenna_metrics attributes return objects that wouldn't be too clear how to use/analyze to a new user.

Other than that, looks great!

tomflexcompute (approved):
Great updates! Thanks @yuanshen-flexcompute

athielens-fc:
@yuanshen-flexcompute, I have added some minor comments below:
- in [2]: add a short comment on top, #importing packages
- In [3]: I would change the text to: "# The Field monitors require separate target frequencies defined here:"
- in [10]: Can the size of the monitor be parameterized in terms of the total simulation size or the size of all structures? Maybe the simulation size should be defined before defining this monitor?
- In [13]: Maybe mention in a comment that we recommend lambda/25?
- In [15]: Perhaps add a comment explaining the choice of run time
- In [25]: $|S_{11}|^{2}$
- In [40]: Could you add more info on why the Axial ratio is clipped?

athielens-fc (approved):
@yuanshen-flexcompute, I have added some minor comments below:

- in [2]: add a short comment on top, #importing packages
- In [3]: I would change the text to: "# The Field monitors require separate target frequencies defined here:"
- in [10]: Can the size of the monitor be parameterized in terms of the total simulation size or the size of all structures? Maybe the simulation size should be defined before defining this monitor?
- In [13]: Maybe mention in a comment that we recommend lambda/25?
- In [15]: Perhaps add a comment explaining the choice of run time
- In [25]: $|S_{11}|^{2}$
- In [40]: Could you add more info on why the Axial ratio is clipped

PR #319, opened 2025-07-01
Differential Stripline Benchmark notebook

yuanshen-flexcompute:
Here's the notebook for the Differential Stripline benchmark.

Unfortunately the notebook suffers from 'WARNING' spam, as the warning messages are not yet optimized for RF simulations. Should I set the logging level to 'ERROR' only?


tomflexcompute (commented):
This is great! Thanks @yuanshen-flexcompute . Some comments mainly regarding formality:

1. Add the notebook to the `docs/case_studies/microwave.rst`.

2. Add the imported files to `misc/import_file_mapping.json` so in the notebook server when user click this example, the imported files will be created in the same folder and they will be able to run the notebook without error.

3. Note that in our convention, in the title (h1 header) only the first letter of the first word is capitalized, while in subsequent headers the first letter of each word is capitalized. Currently they are the opposite. 

4. Maybe add a simple intro in the beginning on differential stripline like what it is, why they are useful etc. Could be good for SEO. 

5. The amount of warnings does seem a bit out of hand so I would suggest we set logging level to error.

FilipeFcp (approved):
Thanks @yuanshen-flexcompute, that is really great.

I would recommend adding links to some relevant classes: TerminalComponentModeler, ImpedanceCalculator, WavePort, etc.

Apart from that, I have nothing to add on top of Tom’s suggestions. I also agree that we should suppress the warnings.

dmarek-flex (changes requested):
Great notebook and comparison! I want to use the same style for plotting errors now. Thanks @yuanshen-flexcompute

1. First paragraph, I would call a stripline a transmission line (instead of a waveguide).
2. Can you add comment for this choice?
`refinement_inside_sim_only=False,`.
If there is no reason to use, remove it. If there is, please briefly explain.
3. I am confused by the plot of the structures. The substrate is transparent, lossy metals are blue, and BCs are gold? I don't think this is something you did, but I guess changes in Tidy3D itself. If not too hard, could you try to add the gold color to all metals, and green/blue to substrate so it matches the other images a bit more closely.
4. Consider plotting the Ex field as well. I think it shows more clearly, along with the voltage integral, that the mode is indeed a differential-mode.
5. I noticed that some plots do not have the y axis labelled. I would recommend adding one with the units. The plot title itself can be optional.
6. Finally, consider plotting S21 in dB using |S21| to indicate absolute value (personal taste :smile: )

yuanshen-flexcompute:
Thanks everyone for comments. 

@dmarek-flex Regarding the last comment about plotting S21 in dB, normally I would do it too. But for benchmarking purposes, I want to show the 0.01 delta S error band, which is more nicely shown in magnitude scale.

dmarek-flex (approved):
Thanks!

PR #320, opened 2025-07-02
Filipe/micro led metadata fix

FilipeFcp:
Hi @alec-flexcompute 
Sorry, forgot to add somethings to the metadata

Thanks!


alec-flexcompute (approved):
Very nice notebook, only a couple small things:

Above cell 7, "Define a function to set up the simulation" -> "Define a function to set up the simulation according to dipole position and its polarization" + a few lines as a precursor to the explanation above cell 13 to explain why we're creating this function in this way

Above cell 14, "averaging the power patterns" -> "summing the power profiles"? I don't see the power sum being divided by anything, so a sum is more accurate. Also pattern is a less familiar word, but it super doesn't matter too much which you use, just a thought

No need to rerun the cells, I think just these description edits are the only necessary edits.

FilipeFcp:
Thanks @alec-flexcompute!

PR #322, opened 2025-07-03
fix[notebook]: remove saved optimization files from autograd notebooks

groberts-flex:
Removes the saved optimization files from repo so that the optimizations run instead of loading old files. If someone runs it locally, they'll create this saved progress file that they can use for checkpointing, but this removes it from being tracked in git.


groberts-flex:
went through autograd notebooks to try and find all of the ones that have checked in optimization histories - think I got all of them!

yaugenst-flex (approved):
LGTM, thanks @groberts-flex!

PR #326, opened 2025-07-04
Grounded CPW with Via Fence

yuanshen-flexcompute:
Here's a notebook of a grounded CPW with via fence. Showcases lumped port excitation, resonance calculation over wide bandwidth (100GHz), and 3D waveguide. Final section compares data with paper results and agrees very well.


alec-flexcompute (approved):
Looks great @yuanshen-flexcompute ! A couple things:

1. It's my preference to have the material parameters and the material definitions in the same cell - something to consider to increase readability slightly. Not necessary to change

2. Throughout the notebook, capitalize "via"

3. I think it might be good practice to label the simulation objects with/out fences to reflect that, instead of the fenceless simulation being labeled "sim"

4. In cells 16 and 25, the x labels overlap. I suggest increasing the figure size

Very nice agreement with Sain et. al!

Also, how did you create the cross-section diagrams below cell 1? They look very nice

yuanshen-flexcompute:
@alec-flexcompute Thanks for feedback! Title figure is created with blender 
image

Also 'via' is not a proper noun and is typically lowercase: https://en.wikipedia.org/wiki/Via_(electronics)

tomflexcompute (approved):
Thanks @yuanshen-flexcompute for the nice example. I didn't have time to look into all the details but everything looks great on a high level. Usually we put the schematic image after the intro paragraph but that's a minor thing.

PR #328, opened 2025-07-14
Edge Feed Patch Antenna Benchmark

yuanshen-flexcompute:
This notebook demonstrates a benchmark simulation of a patch antenna compared with a commercial FEM tool.


tomflexcompute (approved):
Thanks @yuanshen-flexcompute . Great match of the results. A few small typos:

1. "it is also important refine the grid near" -> "it is also important to refine the grid near"
2. "behaviour" -> "behavior"

Frank-DingYC:
Thanks @yuanshen-flexcompute. A few comments on the notebook:
1. Could we change those numbers to variables? It's hard to read and reuse those parameters.
2. Could we use the scripts that are similar to the following style to make the unit easier to read?
```
mm=1000
t = 0.02 * mm
h = 0.254 * mm
```
BTW, here is the paper related to the patch antenna: [[Edge feed, MIMO, 5G].pdf](https://github.com/user-attachments/files/21224146/Edge.feed.MIMO.5G.pdf)

yuanshen-flexcompute:
@Frank-DingYC Interesting, the paper shows that the resonance is ~37 GHz. That is quite different from what I found with Tidy3D and other FEM software (~35.3 GHz). I double checked the dimensions and it's the same. Any idea why? It doesn't look like they said what software they used

alec-flexcompute (approved):
Excellent job @yuanshen-flexcompute! I have very few suggestions, as this notebook is quite clear and well-structured

1. Above cell 13, maybe a brief explanation of why a smaller shutoff is preferred at low frequencies

(Optional) Plot the close up to the corner refinement to show what it's doing. This might be one of the first examples that a user will see, since it's a benchmarking comparison, so more details about our capabilities here are probably preferable

2. Above cell 32, there're two periods

I'm not sure if the section on coarser grid is needed, but it's a good study I think.

Frank-DingYC:
I didn't find the reason either. That's why I created this simulation in both the FEM/FIT tool. The results of those two tools match well with each other, and so does our result. @yuanshen-flexcompute

PR #330, opened 2025-07-16
adding directional coupler example

FilipeFcp:
Hi all,

I added a quick example using the directional coupler GDS, as @e-g-melo suggested.

Today, I noticed that there's a very similar notebook in the [parameter sweep tutorial](https://docs.flexcompute.com/projects/tidy3d/en/latest/faq/docs/parameter-sweep.html), so I focused more on the simulation setup and post-processing in this case.

If you have any suggestions to make it more interesting, feel free to let me know!


alec-flexcompute (approved):
Great example as always, @FilipeFcp! A couple small points:

1. Also mention that you're creating a field monitor to check out the profile in the middle of the waveguides

2. Above cell 9, add "(if the field decays before the runtime, the simulation will auto shut-off and we will only be billed for the Flex credits used to run)" since this will probably be one of the first examples a new user looks at

3. In the interpolation cell, it might be helpful to add a legend in the plot

That's it, thanks for making these examples very easy to review!

e-g-melo (approved):
Thanks for this new notebook @FilipeFcp! It is pretty didactic.

Some suggestions:
1. "and create the `td.Structures` with those material properties" -> "and create the `td.Structure` with each of those material properties".
2. We don't need the plt.show() after sim.plot_3d() in cell [8].
3. I suggest using dB in the chart of cell [11] to display the -3dB crossing point. The current method also works, though. 

Emerson

tomflexcompute (commented):
Thanks @FilipeFcp . Definitely make sense to have this addition in the example library. 

1. The simulation domain size in y can probably be shrunk a bit more to reduce cost?
2. In the plotting, does setting `edgecolor="k"` actually do anything to the plot?
3. Might be better to plot transmission as a function of wavelength instead of frequency.
4. For printing the transmission, might be better to only show something like 4 significant digits?
5. Probably don't need to add the notebook to docs/features/parameter_sweep.rst?

FilipeFcp:
Thank you all.

I believe everything has been addressed.

daquinteroflex:
Let's merge when ready!

tomflexcompute (approved):

PR #334, opened 2025-07-22
Hybrid microstrip-CPW bandpass filter

yuanshen-flexcompute:
A filter model for the RF library. Accuracy benchmark is performed against a commercial FEM tool.


FilipeFcp (approved):
Thanks @yuanshen-flexcompute. Very didactic and easy to follow.

in cell [2]:  
- `Commision` → `Commission`  
- `accounts two` → `accounts for two`    (This is a GPT suggestion)

Other comments:

- Before cell [5], maybe add a quick note on the use case for [`LossyMetalMedium`](https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.LossyMetalMedium.html).
- In cell [7], totally optional, but I like to carry out clip operations with `+` and `-` when possible: `geom_plane - geom_hole`.
- I think it would be good to add links for the classes, such as `LayerRefinementSpec`, `TerminalComponentModeler`, etc.

tomflexcompute (approved):
Thanks @yuanshen-flexcompute sorry for the delay in review. All look great!

HybridMicrostripCPWBandpassFilter.ipynb
dmarek-flex, line 1: Grammar:
"Key settings here are " -> "The key settings here are "

dmarek-flex (approved):
Beauty! One thing to add would be some kind of short conclusion. Like a simple comparison between your results and the reference, even though you compare to other commercial FEM. Maybe some quantity straight from the abstract like the reported bandwidth and center frequency. Or from the conclusion group delay / insertion loss level.

If you have an explanation for the rapid oscillations in the group delay would be nice too. I suspect if you lower the cutoff even more, they will go away?

PR #335, opened 2025-07-22
clarified use of "estimate_cost()" and "real_cost()" in HeatSolver.ipynb

jewettaijfc:
This PR clarifies when `.estimate_cost()` and should be used.

Confusion over this was the direct cause of [tidy3d issue 2551](https://github.com/flexcompute/tidy3d/issues/2551).

Currently, `.estimate_cost()` is invoked immediately after the `job.run()` in the same cell.  Excerpt:
```
job = web.Job(simulation=heat_sim_cyl, task_name="heat_sim_cyl")
estimate_cost = job.estimate_cost()
heat_sim_data_cyl = job.run()
```
Users did not realize that `job.run()` (or `batch.run()`) also invokes `.estimate_cost()`.  So when they saw the cost printed twice, they worried they were getting billed twice.

After this PR, all 3 commands are in different cells, with explanations for each cell.
I also added a line that calculates `.real_cost()` as well after the simulation finishes.

*(Unfortunately, this required re-running the simulations.  I also generated new figures.  So the size of this PR is quite large (844kb), even though I only changed a few lines.  If anyone knows how to avoid this in the future, let me know.)*


FilipeFcp (approved):
Seems great. Thanks @jewettaijfc

FilipeFcp:
I don't think there is a way to not generate the new outputs. But for the notebooks, it is not a problem. Also, it is a good opportunity to double-check if nothing is behaving weird after some update.

jewettaijfc:
Awesome!   Thanks Felipe!

PR #339, opened 2025-07-24
adding GUI script visualization tutorial

FilipeFcp:
Hi @e-g-melo 

This is a first version of the notebook. Let me know if it aligns with what you were expecting.

Just for context, @alec-flexcompute, Emerson asked for a guide for users who want to process multiple simulations from a sweep. Since the GUI currently only allows downloading data for up to 10 simulatons at once, he suggested creating a tutorial to help users work with the script generated by the GUI.

Also, Emerson, I have two suggestions for the script:

    Add a list of the sweep values to make it easier to create labels when plotting the fields.

    Consider including the import statement for matplotlib in the script.


e-g-melo (approved):
Thanks @FilipeFcp! The notebook is amazing!

I will provide the design variable names in the code snippet.

FilipeFcp:
Thanks @e-g-melo.

I forgot to mention. Will the users have access to these task IDs? If I suffice to share the main simulation, or should I share all the sweeps?

alec-flexcompute (approved):
Great job as always @FilipeFcp - I'll probably copy/paste from this notebook myself. A couple quick things:

1. "The good news is that you don't need to be a Python expert." -> "Tidy3D users need not be Python experts to do this." The former way colloquially suggests that there's a "bad news" side to "the good news" stated before.

2. Right above cell one, maybe give a brief (i.e. 1 sentence, nothing technical) explanation of what an "ax" object is.

3. Link to the actual function (https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.components.data.sim_data.AbstractYeeGridSimulationData.html#tidy3d.components.data.sim_data.AbstractYeeGridSimulationData.plot_field) to make it easy to see allowed values for parameters like field_name, val, etc.

4. (OPTIONAL) Above cell 4, you may want to briefly describe and link to numpy's unwrap function (https://numpy.org/doc/stable/reference/generated/numpy.unwrap.html) which is useful for plotting phases.

FilipeFcp:
Thanks @alec-flexcompute! Great catches as always.
I actually never used np.unwrap before, it's so convenient. I don't know how I lived so long without it 😄

tylerflex:
can we merge this?

PR #341, opened 2025-07-25
S Matrix Autograd 

tylerflex:
Requires a merge of frontend PR


yaugenst-flex (commented):
Nice! Here my comments:

1. I know it's topologically the same thing, but this is a very strange looking waveguide crossing 😆 Might be quite confusing to readers to have the ports arranged like that..
2. You actually don't use autograd numpy other than for setting up to (undifferentiated) `linspace`s - I guess it works, which is nice. But I'd be careful and just use autograd numpy throughout?
3. Before Cell 15: Typo: `optimizer.th`
4. Cell 18: You set `beta_min = beta_max = 20`, so no beta scheduling. Is that intentional? If it is, it probably should be commented.
5. Cell 18: `params` get appended to `params_history` and are always mutated in-place, meaning that the final `params_history` will contain the same parameters for each entry. Should be `params_history.append(params.copy())`.
6. `import tidy3d.web as web` unused. Similarly, there are a couple of other formatting issues, should run `ruff format && ruff check --fix`. CI is failing.

tylerflex:
> I know it's topologically the same thing, but this is a very strange looking waveguide crossing 😆 Might be quite confusing to readers to have the ports arranged like that..

can you help me make it look better? how would you change it?

tylerflex:
> You actually don't use autograd numpy other than for setting up to (undifferentiated) linspaces - I guess it works, which is nice. But I'd be careful and just use autograd numpy throughout?

sounds good, fixed

> Before Cell 15: Typo: optimizer.th

fixed

> Cell 18: You set beta_min = beta_max = 20, so no beta scheduling. Is that intentional? If it is, it probably should be commented.

yea it seems to work alright, will add a comment

> Cell 18: params get appended to params_history and are always mutated in-place, meaning that the final params_history will contain the same parameters for each entry. Should be params_history.append(params.copy()).

fixed

> import tidy3d.web as web unused. Similarly, there are a couple of other formatting issues, should run ruff format && ruff check --fix. CI is failing.

thanks, fixed this too

Will change to a "Plus" setup and will re-run but the queue is horrible today so it's like an iteration every 45 minutes..

groberts-flex:
thanks Tyler, this is really cool! A couple of comments below
1. leftover print of the tidy3d version in the import section
2. cell 2: should we remove n_sio2 since it's unused?
3. cell 3: "and a field monitor" -> "add a field monitor"
4. cell 12: some commented code leftover for the `element_mappings`
5. cell 13: looking at all the input/output for the various ports seems useful for debugging, but maybe ok to only plot one of them as an example in the notebook output for conciseness. If possible, it might be good to add a text label to one of them to identify each port index.
6. cell 14: can probably get rid of the printed smatrix output
7. cell 20: leftover print statement in objective

at cell 21, seeing that the output is broken which I'm assuming is due to all the flux in the backend and how the autograd stuff is running. When that part is updated and re-run, I'll add some more comments for the rest of the notebook

tylerflex:
Thanks @groberts-flex , I made all comments and fixed the element mappings too so it only solves from one port (since all results can just be rotated through the ports). 

Sorry you saw this version, I had made some major changes after Yannick's initial reviews so it was a bit messy. but appreciate the help cleaning it up.

Once things are back up and running I will polish up and get things into final state.

yaugenst-flex:
oh another comment i just remember: we should use `td.Structure.from_permittivity_array(eps_data=eps_data, geometry=design_region_geometry)` instead of the custom medium + spatial data array coordinate wrangling

Autograd26Smatrix.ipynb
yaugenst-flex, line 1: **Markdown above [9]**: "parmeterization"
**[9]**: `functools.partial` is unused now right?
**[14]**: Nit, but the `dtype=float` isn't necessary. It'll be promoted in the objective function, so not including this might improve readability a bit.
**[15]**: Remove commented out import.
**[17]**: Lots of `ipywidgets` warnings. Might be nice to just install that locally when running the notebook so it doesn't clutter the output here?
**[26]**: Any reason not to plot this over wavelength? Currently seems somewhat difficult to parse, especially since it's plotting normalized frequency. Also, a before/after plot would be nice..

tylerflex:
Thanks @yaugenst-flex , all suggestions have been addressed. notebook was re-run and for some reason the grey region just went away by itself. Squashed commits. Should be ok to merge now @daquinteroflex

tylerflex:
This PR will merge into pre/2.10 branch btw

PR #344, opened 2025-07-28
Further notebooks rerun

daquinteroflex:


daquinteroflex:
They're good @yaugenst-flex

PR #345, opened 2025-07-28
using fsolve to find L_max

FilipeFcp:
Hi all,

I am proposing a small modification to the calculation of `L_max`. This is the second time a user has reached out after changing a parameter and either getting stuck while finding `L_max` or receiving an unexpected value. With this change, the cell won’t get stuck if `L_max` passes through the smallest value without converging, and it will also return a message if there is no solution for the given A.

I tried to find an easy way to calculate the maximum and minimum A for a given R_eff, but couldn't figure out a straightforward method.


tomflexcompute (approved):
Looks good to me. Thanks for the fix @FilipeFcp !

momchil-flex (approved):
Nice, indeed this looks much better. `while True:` rings all sorts of alarm bells.

PR #346, opened 2025-07-31
adding path arg to avoid potential name colisions

FilipeFcp:
Hi @tomflexcompute

An user from Lumotive said he was having issues with the final part of the [parameter scan](https://docs.flexcompute.com/projects/tidy3d/en/v2.0.2/notebooks/ParameterScan.html) notebook, probably due to a name collision.

I think it can’t hurt to set a different path for each simulation, just to be safe.


tomflexcompute (approved):
Looks great thanks @FilipeFcp

PR #352, opened 2025-08-11
Added quasi-gaussian beam to AdvancedGaussianSources

alec-flexcompute:
added cells 2-6 and in the other cells corrected some variable names


FilipeFcp (approved):
Thanks @alec-flexcompute.  
It looks great.

I have just two totally optional comments:

The interactive widgets are really cool, but the images can only be viewed by running the notebook. So maybe it would be nice to add a plot like the one you included in cell [6] for the other sources.

Also, it seems there is some redundancy in the code, so maybe a method `.to_source` could be handy to group these lines:

```python
    field_data = Bessel_obj.field_data
    field_dataset = td.FieldDataset(
        Ex=field_data.Ex,
        Ey=field_data.Ey,
        Ez=field_data.Ez,
        Hx=field_data.Hx,
        Hy=field_data.Hy,
        Hz=field_data.Hz
    )
    Bessel_source = td.CustomFieldSource(
        center=center,
        size=size,
        source_time=source_time,
        field_dataset=field_dataset
    )

alec-flexcompute:
Thanks @FilipeFcp! For the first suggestion, I'll hold off on that for now, as I'm going to be adding thin lens hopefully soon and Ince-Gaussian beams as well.

As for the other method proposal, the idea is that a user would need only copy the single cell in order to use the source they need, so I don't actually want to reuse methods between these sources, as this would be less copy-able

tomflexcompute (approved):
Thanks @alec-flexcompute . Not in this notebook but it would be good for internal benchmarking purposes to compare the quasi-gaussian and our paraxial gaussian beams in terms of the accuracy of the beam waist, focal length, etc to see how much advantage the quasi-gaussian beam has when the waist is small.

alec-flexcompute:
Good idea @tomflexcompute, I'll do that tomorrow

momchil-flex (commented):
Addition looks good!

I think Filipe's note about the widgets is valid, now that you have removed the direct plots, the notebook / docs page will look like just a lot of code without any viz? Not sure why you decided to remove all plots?

>  I'm going to be adding thin lens hopefully soon

Are you planning this as a notebook addition or a core functionality?

alec-flexcompute:
Ah, I looked at the old version and realized I forgot that it had 2D plots that were display-able. Not sure why they were removed in the version I added this, but added them back it.

@momchil-flex I was assuming it'd be a notebook addition but could make it a core functionality as well. Let me know if you have a preference.

PR #355, opened 2025-08-18
Add PIN Mach-Zehnder notebook

damiano-flex:
This pull request contains a draft version of the PIN Mach-Zehnder notebook.

To better demonstrate the model's capabilities, the electrode temperature has been artificially increased by 50 degrees. This change is intended to showcase the different simulation outcomes between an isothermal analysis, a non-isothermal analysis, and one that accounts for the temperature-dependent effective density of states (DOS).


alec-flexcompute (approved):
Great example @damiano-flex ! A couple notes for publishing the notebook on our website's example conventions:

1. Section headings should have only the first letter capitalized, where usb-headings should have every word capitalized.

2. Below cell 3, "Mobility models Models" -> "Mobility models"

3. Below cell 11, maybe describe around where Fermi Dirac statistics become the more accurate choice

4. Below cell 12, "in particular, an higher refinement..." -> "in particular, a higher refinement..."

5. When running the simulation in cell 16, the logging is quite verbose and should be suppressed

6. In cell 21, I don't think "Eff. DOS - CST" is necessary as a label, but if you want to include it, maybe remind the user what "CST" is

Other than that, looks very informative and useful to users!

momchil-flex:
Thanks @damiano-flex!


A few initial comments from me:

- The notebook title should not be Introduction but the actual title you want it to appear under in our documentation.
- "The operation of the PIN diode generates heat, especially at high modulation speeds and power levels." A few questions about this. First of all, I'm wondering why the modulation speed will matter. If anything, I would think that the highest heating will happen when the modulator is constantly "ON" (current flowing), while if modulation is happening, some averaged heating between the "ON" and "OFF" state will occur. Regarding "power levels", are you referring to the optical power? Because indeed that could lead to heating but it's the sort of heating that is completely not included in the simulations in the notebook.
- I think the bottom two plots in cell 9 are kind of confusing. I would suggest just leaving the top two ones, I am not sure what exactly you're trying to convey with the bottom ones.
- "Isotropic Effective Density of States (DOS): A novel feature, IsotropicEffectiveDOS": let's avoid commenting on the novelty. You should think of this notebook as something that will live in our documentation for a while, so at some point it will no longer be "novel" and this comment becomes stale. :) *If* needed, you could say something like "IsotropicEffectiveDOS (introduced in version 2.10) ..." but I think there's no need here?
- The local refinement image is not showing for me in cell 12
- `getInterpolatedData` - normally we use `snake_case` across our entire python codebase. Similarly, this strikes me as something I wouldn't do: `XI, YI = np.meshgrid(xi, yi)`. Variables should be lowercase unless there's a very good reason for them not to be.
- Probably already right after the introduction you should explain that we'll be doing three simulations and what the difference between them will be, since now there are several places where these differences are defined and an initial overview would be helpful.


image

marc-flex:
Thanks Damiano for preparing this! 
I think my main comment is that I think you should explain what you intend to do before doing it. In this case, you're creating 3 different simulations. I think it should be explained at the beginning why you're going to be creating these and what you expect to see with these. (it's kind of obvious to us but we should be explicit)

Some more concrete comments:

- In the introduction I'd remove the point on carrier mobility. Not only mobilities are temperature dependent but other properties are as well. So I'd say instead that the semiconductor behavior is very sensitive to temperature.
- When talking about the mediums, maybe change this `This is an alternative to using a constant value.` to something like `This is a temperature dependent model.`
- Also maybe before defining the mediums say something like you're going to compare results between a case that has temperature dependent DOS and cases that doesn't so that users aren't surprised when they see things like `ConstantEffectiveDOS`
- In boundary conditions, you're using both `HeatChargeBoundarySpec` and `HeatBoundarySpec`. The latter is deprecated and should be replaced by the former.
- In general I think you should mention what you're going to do before doing it. In particular, when the simulation objects are created you could maybe explain what are the differences between the 3 simulations you're creating.
- Some comments about the results would be nice. For instance, you're plotting the potential field for different bias and the differences with different cases. Maybe comment what differences you're appreciating. To be fair, though, I'm not sure these fields add much value to the user except that for at 1.2V bias, you can see a ~linear variation of potential indicating conduction. So maybe with just a plot showing the potential field at two applied bias is sufficient. I'll leave for the team to decide
- A similar thing could be said about the temperature plots though in this case it's maybe more interesting. I'd add more description of what we're seeing
- A little explanation about IV curve behavior would also be nice

marc-flex:
This is great! The only thing that I'm missing is a little analysis/description of results (I added some suggestions down below). Other than that just a few minor comments. 

- `This specific version of the notebook...`: I'm a bit confused by this sentence. Is there another version of it? If there isn't maybe something along the lines of `For this notebook we make use of different models for the effective Density of States (DOS)...`
- In section "Scope" you're saying that we're going to demonstrate how to use multiphysics tools to perform a comprehensive simulation of PIN Mach-Zehnder but in reality we're only assessing electric and thermal characteristics, right? So I'd say maybe best to say something like `This notebook demonstrates how to use our multiphysics simulation capabilities to assess the thermal and electric characteristics of PIN Mach-Zehnder modulator which are essential for the operation of such device.`. What do you think?
- subprocess, h5py, pandas and LogNorm can be removed from the notebook? Eventually also tidy3d_backend. EDIT: I just saw a TODO below where you mention you need to remove some code which seems related to this.
- In the doping section: the source parameter isn't the axis along which the Gaussina function decays. It decays along every direction. The source indicates the side of the box where it doesn't decay
- In nwell and pwell, in size you have y dimension 1.0001. Is this to provide some kind of overlap? If so, it may be good to say it somewhere.
- In the "Media".  You're saying air and contact media are defined as fluid to remove them from the thermal simulation. The contacts don't need to be fluids to be removed from the thermal simulation. In principle, they could be any type of media as long as they are charge conductors 
- In "Mobility models": `we use several advanced physical models`. I'd avoid calling them advanced. 
- In the doping plots, it may be worth zooming in near the waveguide for the plot on the right? I leave it at your discretion 
- In "Simulation Setup": in point "Grid Specification" the last "GridRefinementRegion" it's probably better to have it enclosed in `
- In "Post-processing and Plotting": `How the potential is distributed in the simulation region.` -> `Electric potential field distribution in the simulation region.`
- A little comment/analysis of the plots. For instance in the plots of cell 29 and 30 you could say stuff along the lines: "As it can be seen using this and that models, produces an increase/decrease in potential. This is due to this and that effect on DOS which in turn produces a change in equilibrium densities"
- In cell 37 also some explanation of what we're seeing would be nice. Also, a little analysis of things are changing the way they are?
- Similar thing for the temperature plots. Maybe explain that the temperature with certain model is higher because of something (maybe the model increases densities which in turn increase currents which increase heat creation?). Maybe also mention that to better appreciate the temperature differences you plot the difference distribution which highlights how temperature where the higher currents are is where most differences are

momchil-flex:
Thanks @damiano-flex looks great. Just some small things:

- I would suggest adding "forward-bias" in the title explicitly. On the other hand, you don't really demonstrate an MZM: you just discuss the phase shifter component (unlike [here](https://www.flexcompute.com/tidy3d/examples/notebooks/MachZehnderModulator/), where we show the actual MZM). So how about "Forward-bias PIN phase shifter with thermal analysis"? I think you can still mention that this can be used in an MZM but probably more accurate later on in the text too to just call it a phase shifter rather than an MZM.
- There's still some things that need to be removed, I think you're aware of that?
- My usual comment for plots to not use width as large as 14. 12, or even 10 instead should look better (although you do have some long titles in some of them... maybe they can be shortened, or they can stay at 14)

marc-flex (approved):
This looks great @damiano-flex 
I added some minor comments but it's good to go

These seem to be unused
```python
import subprocess
import pandas as pd
import h5py
```

I would eliminate these sentences:
```markdown
- **P-Side (right):** Requires **more** bending, creating a positive potential difference.
- **N-Side (left):** Requires **less** bending, creating a negative potential difference.
```
or change them to something like
```markdown
- **P-Side (right):** Ei bends towards the valence band creating a positive potential difference.
- **N-Side (left):** Ei bends towards the conduction band creating a negative potential difference.
```

I would eliminate section "Comparative plot"

In the sentence
```
The plots above highligh how the **Constant DOS** model underestimates the number of carrier interactions at high temperatures
``` 
highligh -> highlight.
Also it is a bit generic (sounds AI generated). Instead I'd say that since the effective DOS are allowed to change more the resulting increased intrinsic concentration leads to larger currents which result in larger generated heat.

damiano-flex:
I updated the notebook:
-  Addressed the last comments.
- Removed the code parts required to run it locally
- Made a run on servers.
- Adjusted the x sizes to 10, while keeping 12 only when necessary.
- Changed the title to **Forward-bias PIN phase shifter with thermal analysis**, and also slightly updated the text to match more consistently with the new title.
- updated the metadata.

The only thing I am missing is a feature picture. Maybe @tomflexcompute can produce it?

edit: I am giving it a try with the picture @tomflexcompute

momchil-flex (approved):

PR #356, opened 2025-08-20
fix: file path in import_file_mapping.json

huwei-mc:


FilipeFcp (approved):
My bad, the json is already in the misc folder.
Thanks for fixing it @huwei-mc.

PR #361, opened 2025-08-25
Microwave frequency selective surface to use lossy metal and layer refinement spec

weiliangjin2021:
This is brought up by @groberts-flex: this notebook is using Medium2D rather than lossy metal to model copper. Additionally, I upgrade mesher to use layer refinement, and you can find much fewer grids are used now.

Side note: in upgraded singularity correction, a closer result to the original notebook can be found as attached. I'll update the figure once that change is merged.
image


tomflexcompute (approved):

groberts-flex (approved):
thank you for updating!

PR #363, opened 2025-08-28
changing td.inf to effective inf

FilipeFcp:
Hi all. 
Just updating the notebook without the `td.inf`.
If we fix that, we can revert to the previous version, but I think it is good to have a working version in the library.


momchil-flex (approved):
On a separate note, `td.inf` has been a source of headaches in some cases.

We had a related discussion with @dmarek-flex about scales in the RF solver, where it seemed that e.g. CST does not allow to define objects larger than several orders of magnitude larger than the simulation domain, or something like that.

Should we only support an "effective infinity" as `td.inf`, e.g. setting it to 1e7? We already do some conversion like that at least in some places I believe. @lucas-flexcompute @tylerflex any thoughts?

tylerflex:
> Should we only support an "effective infinity" as td.inf, e.g. setting it to 1e7? We already do some conversion like that at least in some places I believe. @lucas-flexcompute @tylerflex any thoughts?

I'd prefer keeping `td.inf` as something that has a specific meaning as "extends to infinity" and only set it to a value where absolutely needed. Otherwise I can imagine things getting quite confusing real fast, although I can't point to a specific case yet.  Well, maybe one example would be if setting something eg `Box.from_bounds()`, if we set rmin and rmax to ±td.inf respectively, you could imagine a situation where we get some weird rounding error where the center is not exactly zero. I think we should handle these kind of cases checking explicitly if `td.inf` is passed.

This is mostly just my gut feeling though. maybe others have stronger opinions.

momchil-flex:
The point is that it's too much freedom for users and quite a pain to handle everywhere under the hood. For example the issue here was that geometries with infinities can't be Transform-ed. So we can't use transformations under the hood when we need to, or we have to figure out what to do with geometries that have infinities defined. This pops up every now and then in general.

tylerflex:
> The point is that it's too much freedom for users 

what do you mean?  I dont think this is in the scope of this issue because it's just discussing a backend change right? but I'd think that using `td.inf` is convenient for users as opposed to setting a large number.

>  For example the issue here was that geometries with infinities can't be Transform-ed. So we can't use transformations under the hood when we need to, or we have to figure out what to do with geometries that have infinities defined.

yea so my opinion is it would be better to figure out what to do with geometries that have infinities defined rather than just set all infinities to a specific value, which could cause more issues down the road.

tylerflex:
For a very simple example, if I have a `Cylinder` with a `length=td.inf`, it would be nice if the `td.inf` was retained under some transformation. I think it would be confusing and potentially cause issues if the transformed cylinder now had a length of 1e7 or something

marc-flex (approved):

lucas-flexcompute:
My first though is to agree with Tyler that having `inf` as a exactly infinite is preferred in general, because is clearly denotes the intent. I also understand how that can be annoying when doing any transformation (and that because of that we do have some work-around code throughout our whole geometry kernel), but I still think it is better this way. Whenever operations support handling `inf`, that's a good thing.

In photonforge we don't really have this option (at least not automatically) because we use integer coordinates internally, so there's no `inf` representation available unfortunately.

PR #372, opened 2025-09-11
adding SiO2 medium to optical simulation

FilipeFcp:
Hi @marc-flex. I added the SiO2 to the optical simulation and added a line to estimate the order of magnitude of the charge results.

In the future, do you think it would be possible to make that internally at the `perturbed_copy` function, so the user doesn't need to take care of it?


marc-flex (approved):
Thanks @FilipeFcp for updating this. Only one question though: this notebook needs to be run, right? I don't see any output

FilipeFcp:
Sorry, I cleaned the results before saving it. Now should be all there

marc-flex:
The real part of the refractive index looks now different but the normalized transmission power looks much better!

PR #373, opened 2025-09-12
remove: adjoint plugin

daquinteroflex:


tylerflex:
I thought it would be sufficient to remove the link to adjoint plugin in the main TOC, but keep the notebooks in case people need them. 

But now that I think about it, maybe removing is better and then we can refer people to old versions of the docs if they want to see them?

daquinteroflex:
Yeah the old versions of the notebooks are on the docs so this is only for 2.10+ anyways? It simply won't work for 2.10

tylerflex:
yea sounds good

tylerflex (approved):

yaugenst-flex (approved):

PR #375, opened 2025-09-12
Through Silicon Via

yuanshen-flexcompute:
Example showing how to model a through-silicon via


tomflexcompute (approved):
Very nice notebook @yuanshen-flexcompute . The only drawback is that it's hard to visualize the 3D geometries and users need to mentally imagine how they look like as they are created.

alec-flexcompute (approved):
Excellent job @yuanshen-flexcompute! I actually don't have any feedback - it all looks good and clear. One thing I'm wondering - is this based on a paper? How did the design parameters get decided?

yuanshen-flexcompute:
@alec-flexcompute I think some of the dimensions (like layer thicknesses) were based on the paper (or other similar papers), and other dimensions I just estimated.

PR #377, opened 2025-09-17
fix capacity units

FilipeFcp:
Hi @marc-flex 

Just changed the capacity units and updated the source and `HeatChargeSimulation` object in one of the examples to avoid the warnings.


marc-flex (approved):
Awesome! Thank you @FilipeFcp !

PR #378, opened 2025-09-18
uploading CPO heat example

FilipeFcp:
I believe the results look like the paper, although I had to make some guesses on the material properties and heat power. 
@momchil-flex, feel free to take a look if you are interested.


docs/features/heat.rst
```diff
@@ -8,3 +8,4 @@ This section introduces the HEAT solver, a Tidy3D feature that solves the steady
     :maxdepth: 1
 
     ../../HeatSolver
+    ../../CPOHeat
```
momchil-flex, line 11: Isn't the usual policy for one notebook to go in one place only?

I'm not necessarily opposed to linking in multiple places just wondering.

tomflexcompute (commented):
Great result @FilipeFcp and showcase of transient thermal analysis! 

1. I suggest a more formal title such as "Steady-state and transient thermal analysis for co-packaged optics"

2. Since there are quite a lot of warnings, explain why they are and why it's fine to ignore in this case?

3. In the temperature plot, plot the structure boundaries as well?

tomflexcompute:
Also tagging @prashkh since he is interested in this model as well.

momchil-flex (approved):
Looks great and seems to match the paper reasonably well! I wonder if the difference could be coming from the insulating BCs along x and y, in reality there's probably some heat going out of there too?

Just one note: the schematic appears on the top of the first cell but at the end of that cell it says "A schematic of the simulated structure can be seen in the image below:"

Just out of curiosity I noted that the meshing and solver seem to take about the same time ~25 minutes each. Hopefully we'll improve on the meshing side!

image

alec-flexcompute (approved):
Excellent work as always @FilipeFcp! A couple suggestions:
1. You might want to define the unstructured mesh in terms of one of the feature variables. Raw numbers might be more unclear as to why they were chosen.

2. For the two time_specs, you might want to deifne the time_step and total-time_steps variables at the beginning of cell 71 to have it match cell 79

marc-flex (approved):
Thanks @FilipeFcp. This is an awesome notebook! 

Is it possible to use `MultiPhysicsMedium` instead of `Medium` to run this notebook? I feel like it would make sense to use it.

prashkh:
Very nice notebook! I am sure a lot of people will be interested in this work. Thanks @FilipeFcp !!

FilipeFcp:
Thanks everyone for the comments. Sorry, it took me a while to implement; I was running some mesh tests, and the 2D transient simulation takes quite a long time to run.

With a coarser mesh, the 2D transient simulation shows some strange dips, so I kept the first one.

@momchil-flex, these are the boundaries that they report in the paper. I think the discrepancies are in the source definition. They won't report in volumetric power, and I didn't find it in the paper any information to directly translate it.

PR #379, opened 2025-09-19
RF electrode in a microring resonator

yuanshen-flexcompute:
Notebook demonstrating RF electrode in a microring resonator.


alec-flexcompute (approved):
Nice job as always @yuanshen-flexcompute ! A couple quick things:

1. Above cell 6, "The geometry are combined..." -> "The geometries are combined..."

2. After cell 15, the cells haven't been run, so none of the results have been plotted.

3. A bit more description on the lumped RC element, and how this approximates the p-n junction. I found this a bit difficult to understand. A brief description of the geometry would be enough help for my non-RF-experienced background.

tomflexcompute (approved):
Thanks @yuanshen-flexcompute that's great. Like Alec pointed out, the result sections are not ran so we don't see how the result looks like but I'm sure they look fine.

yuanshen-flexcompute:
Thanks for the comments! I agree that the lumped RC calculation could use more explanation, but I didn't feel like I understood it well enough to give a thorough explanation. Instead I ended up referring the reader to 2.2.3 in [1] where it is presented in detail.

PR #381, opened 2025-09-23
Planar Helical Antenna Array

yuanshen-flexcompute:
A notebook demonstrating a full antenna array simulation of planar helical design. Key teaching focus includes how to generate radiation plots for different port feed amplitude/phase.


athielens-fc (commented):
Hi Yuanshen,

I love the notebook. This will be super instructive and I am sure many people will use it to analyze their arrays :)

My comments are the following:

-Intro:
“Helical antennas enable easier polarization matching, and thus a stronger received signal, without the need for potentially complicated and lossy polarization switching circuits. The phased array layout further expands the coverage angle of the device.”

I suggest to change it to:

““Helical antennas enable easier polarization matching, and thus a stronger received signal, than a combination of linearly polarized antennas, which typically need potentially complicated and lossy polarization switching circuits.

Helical antennas are typically 3D structures. However, planar designs using via’s have been proposed in literature. This notebook simulates such a quasi-planar design.

One downside of the design is that it has a limited coverage angle. A potential way of dealing with this is to use a phased array of planar helical antennas, which expands the coverage angle of the device. This array is demonstrated here”

-Key Parameters:
It could be useful to show a top view of one element before you define all the parameters, it would make it easier to understand

-I would delete: “which will suffice for demonstration purposes.”
-In [4]: results in “Output()”, which seems strange
-In [5]: it was to me not intuitively obvious what the “rounded” variable does. Perhaps add some context as a comment or in the text above the cell
-In [5]: I would clarify in the text above that you do not only return the trace geometry, but also the end point, which you need to create the next turn of the helix
-In [7] and [8]: I would reiterate with some short comments that the rboxes will be used for mesh refinement (similar to what you did in front of [9] where the array is created).
-I would put the array creation in a function too, since you used a function on all the other levels.

In [13]: I would say something about the minimal number of steps per wavelength
-In [19]: Do you know why it seems that the metal layer is in the middle of the substrate here?
-Under S-parameters: Can you provide more information on how the ports are numbered (which one is 1 , which one is N_ant)? Is this numbering from [0,N_ant-1] or [1,N_ant]? (I think it is the first one, but you use a conversion to the 2nd?)
-In the Sii plots: axis labels “S_{ii} [dB]” and “Frequency [GHz]”
-For the isolation, my suggestion is to plot a matrix of Sij at the target frequency (colums and rows are the antenna element) with a colorbar, then you can immediately see all the cross port S parameters at the frequency of interest
-To demonstrate the beamsteering: I would definitely include the zero phase shift case and show symmetrical values (-90,-60,-45,-30,-15,0,15,30,45,60,90). I would consider limiting the bottom to -10 dB, that way you will get a much nicer detail on the beams
-I would add something in terms of sidelobe amplitude versus the main lobe
-Some small example on how to characterize such a lobe (direction, 3dB angular bandwidth, energy/power in the main beam, maybe something else? ) would be good
-The scan pattern is awesome

tomflexcompute (approved):
Thanks @yuanshen-flexcompute for another great example. I would suggest adding x and y labels in the S parameter plots.

yuanshen-flexcompute:
@athielens-fc The notebook is updated. Thanks again for the feedback.

PR #383, opened 2025-09-25
update: add spline interpolation to autograd taper notebook

bzhangflex:
Adds interpolate_spline to the autograd taper notebook. Please let me know what you think @yaugenst-flex


bzhangflex:
Instead of adding bloat to the original taper notebook I created a new notebook and linked them to each other.

PR #384, opened 2025-09-30
added transient thermo-optic shifter example

alec-flexcompute:


tomflexcompute:
@alec-flexcompute the schematic image file is not uploaded so it doesn't show up in the notebook.

FilipeFcp (approved):
@Thanks Alec. This is great. The slides showing the fields at different time steps are pretty cool. I don’t have anything to add.

tomflexcompute (approved):
Thanks @alec-flexcompute good to go after checking the following

1. In the notebook the image still doesn't show. Is it just a web display issue?
2. Please edit the active pic component rst file.
3. Before merging, please make sure metadata are added correctly.

alec-flexcompute:
Thanks for those checks @tomflexcompute ! They should be addressed in this latest push

tomflexcompute (approved):
Thanks @alec-flexcompute should be good to go. I think usually we can only have one rst file edited. Maybe just the pic_active.rst in this case. In the notebook metadata, the applications field can have multiple items.

alec-flexcompute:
Sounds good, I think it's only added to the pic active rst. I also added "Heat" to the applications metadata and will merge

PR #385, opened 2025-10-01
LED LEE calculation example

FilipeFcp:
Hi all. 
This is the LED notebook, based on [this](https://ieeexplore.ieee.org/document/9019847) paper. It was a bit hard to match the results, but I noticed that they match well around 470 nm. The authors don't specify the wavelength; they just disclose the source frequencies. 
Anyway, analysing at 470 nm, the results match reasonably well. The issue with this notebook is that it is expensive. A few hundred FCs. 
Do you think it would be a good idea to let the run functions commented, so none will run it by mistake?


alec-flexcompute (commented):
Great job on a huge and complicated notebook! I think it'd probably be a good idea to comment out the run lines as well as include a note that they're commented out underneath the flex credit warning. A few additional comments:

1. The title is missing

2. In cell 3, you might want to specify the "cm^-1" comment

3. Above cell 20, remove comma after lateral.

4. In cell 30, maybe a brief description of what the input (lee_dict) will be

5. In cell 23, consider possibly removing the "#" in the print statement

6. In cell 31, I don't see the error decreasing with size, although you could still conclude that a single dipole is a good approximation. Maybe I'm missing something!

tomflexcompute (commented):
Thanks @FilipeFcp for working on this great notebook. Besides what Alec mentioned, 

1. Above cell[4], "we will decrease the shutoff parameter from the default value of 1e-5 to 1e-4." It seems like it's increasing the shutoff instead of decreasing?

2. In cell[9], might need to do `batch_varying_abs.estimate_cost();` to avoid printing out the cost number as a separate string. Same for cell[12].

3. This notebook is added to the `parameter_sweep.rst` so it will appear as a Python feature tutorial but I think it's more suitable for the example library like the other LED example?

FilipeFcp:
Thank you all.  
I believe everything is now addressed: I added function docstrings and included another data point in the last result to make the trend clearer.  

Regarding the metadata, @tomflexcompute, this one is similar to the Blue µ-LED example.  

For this notebook:  
- `applications = ['Nanophotonics']`  
- `features = ['Parameter sweep']`  

For the Blue µ-LED:  
- `applications = ['Nanophotonics']`  
- `features = ['Parameter sweep', 'Far field projection']`  

Do you think I should change them?

tomflexcompute (approved):

PR #386, opened 2025-10-06
2x2 MMI Power Splitter v0.

e-g-melo:
Hi! This is a kind of experiment using the MCP. The whole notebook was created using Cursor and the MCP. I have not added or modified any code, markdowns, or comments by hand (except for adding the image and metadata). Let's see how it performs :)


alec-flexcompute (commented):
Interesting results @e-g-melo! Flex agent did a great job!

- In the intro, the note on divergences seems a bit out of place - is very chatGPT-y

- The cell where it builds all structures, all simulation components, and the simulation itself is interesting. I guess since it's all in one cell this was all one successful prompt? If so, I'm curious if it could be trained to break up all the code it generates into different sections

- Under key performance metrics, it looks like some redundancy appears in the lists, a classic llm listing property. Power splitting ratio and splitting imbalance are basically the same thing. That being said, the two plots for these values are nice to have. Did you deliberately tell the LLM to plot both power splitting and splitting imbalance?

- The try-exceptions of the plot are again classic llm, probably not necessary in a notebook for tasks like these

e-g-melo:
> * The cell where it builds all structures, all simulation components, and the simulation itself is interesting. I guess since it's all in one cell this was all one successful prompt? If so, I'm curious if it could be trained to break up all the code it generates into different sections

Yes, usually the assistant organizes the setup into different cells, even for a single prompt. I think we can make it more explicit in the rules.

e-g-melo:
> * Under key performance metrics, it looks like some redundancy appears in the lists, a classic llm listing property. Power splitting ratio and splitting imbalance are basically the same thing. That being said, the two plots for these values are nice to have. Did you deliberately tell the LLM to plot both power splitting and splitting imbalance?

No, the assistant has suggested these metrics and created all the post-processing code. In the final notebook version, we can simplify it.

Thanks for your comments and suggestions! I will address them.

FilipeFcp:
Thanks @e-g-melo!
This is actually pretty cool. The MCP did a great job, especially in making the workflow clear and didactic.  

I am just curious about the term *“Professional cosine bends.”* I don’t recall ever seeing a *“Professional cosine”* before, but maybe I’m just not familiar with this terminology.  

Also, in the post-processing, I think the MCP didn’t account for the fact that the sources are normalized. The definition of *power efficiency* should be $P_{out}/P_{in}$ and not $P_{out}/max(P_{out})$, correct? If so, it is simply the total transmitted power.

FilipeFcp:
Also, did the MCP figure out the optimized simulation size alone? Because it is pretty nice if it did.

tomflexcompute (commented):
Very interesting experiment @e-g-melo ! Maybe for the hype you can add a note in the begining of the notebook that something like "this notebook is 95% AI-generated." A few comments:

1. In the markdown it mentions "wavelength sweep analysis" twice. Probably not an accurate description since FDTD is broadband. 
2. In the introduction it mentions MMI has broadband performance. However the equal splitting ratio doesn't seem to be very broadband?
3. As Filipe pointed out, "professional s-bend" sounds quite strange. Also "high-quality s-bend" sounds a bit strange too.
4. AI coding really likes try except, which I suppose is fine.

e-g-melo:
> Also, did the MCP figure out the optimized simulation size alone? Because it is pretty nice if it did.

Hi @FilipeFcp! Thanks for your comments! You are right, I have excluded some of those plots. 

Yes, the MCP has suggested all the initial parameters. Really very nice

e-g-melo:
Thanks @tomflexcompute! AI is really quite creative!! I have addressed all your comments.

e-g-melo:
@tomflexcompute, @FilipeFcp, @alec-flexcompute, Thanks for all the comments and suggestions. Please let me know if you have more comments.

tomflexcompute (approved):
Thanks @e-g-melo all look good to go. I didn't check the notebook metadata but as long as it's good, feel free to merge it.

PR #388, opened 2025-10-09
Updated the unsteady heat notebook metadata

tomflexcompute:
Current metadata generates duplicated entries at the learning center so I modified it a little bit.


alec-flexcompute (approved):
Good catch!

PR #390, opened 2025-10-13
upload phase calculation example

FilipeFcp:
Hi all. A user asked for an example of how to calculate phase. I had this a long time ago as part of the vortex metasurface notebook .
I didn’t include it in the notebook, but I thought it would be a good idea to make a simple example for it.

Almost all comments in the code are AI-generated


tomflexcompute (commented):
Thanks @FilipeFcp . This is the prelude of the vortex beam generation notebook you created. 

In this notebook you used PML on x and y and used the field monitor to extract the phase. I wonder if it's more common to use periodic boundary (local periodic approximation) and diffraction monitor? Is there a particular reason why you chose to use pml? I suspect both methods will give you very similar results.

alec-flexcompute (approved):
Nice job @filipe! I have just two suggestions:

1. In cell 2, maybe specify what the "hemisphere" parameter does in the function

2. Maybe plotting one of the simulations before running so we can visualize the field and source setup

FilipeFcp:
@tomflexcompute 
Yes, using the `DiffractionMonitor` and PBC yields the same results (with a smaller computational domain).
Do you think I should keep both approaches for didactic reasons, or just use the DiffractionMonitor approach?

tomflexcompute:
> @tomflexcompute Yes, using the `DiffractionMonitor` and PBC yields the same results (with a smaller computational domain). Do you think I should keep both approaches for didactic reasons, or just use the DiffractionMonitor approach?

Yes I think showing both approaches would be great.

FilipeFcp:
Thanks, everyone.

I’ve made the changes based on Alec’s comments and added the section using the `DiffractionMonitor`.

Thanks!

PR #392, opened 2025-10-15
feat: add SSAC notebook

damiano-flex:
This notebook is a simple SSAC example. It is straight to the point, without excessive explanations, since it will be quite close to the PIN notebook in terms of new functionalities (other than SSAC).


alec-flexcompute (approved):
Nice job @damiano-flex! This is a very informative example. A couple comments:

1. In the library import cell, you have a "# TODO REMOVE" comment there. I'm guessing you should remove the run_heat import?

2. Same thing in the cell after "Run the Simulation" - remove the "TOREMOVE" comment

3. When discussing the physical interpretation of results, remind us what W is.

4. In the conclusion, "Desigend -> Designed"

5. Please reset the kernel and rerun all cells so the cell numbers are in order.

damiano-flex:
Thank you @alec-flexcompute for the prompt review.

1-2-5) Since these simulations are quite expensive to run, I wait till the last moment to run them in the servers. So I keep the version with the local runner until it is ready to merge.

3-4) Ops. I'll take care of that.

alec-flexcompute:
Also should ruff this notebook so it will pass the lint checks

momchil-flex:
Thanks @damiano-flex ! Ran this by Gemini and it has an interesting suggestion: compare to DC theoretical capacitance

```
# Theoretical capacitance calculation
eps0 = 8.854e-18  # F/µm
eps_si = 11.7
area = device_width * device_depth  # µm^2
width = i_thickness  # µm

C_theoretical = (eps0 * eps_si * area) / width  # in Farads
C_theoretical_pF = C_theoretical * 1e12  # in pF

# Add this to your capacitance plot
ax2.axhline(y=C_theoretical_pF, color='k', linestyle='--', label=f'Theoretical C ({C_theoretical_pF:.2f} pF)')
ax2.legend()
```

Some other comments:
- I would suggest using max width of 10 for the plots (e.g. figsize=(10, ...)). Maybe a bit of a personal preference but I find that they get too stretched and the tick labels get too small.
- Maybe discuss this a bit more in the markdown and not just in the comment. Like what units `ac_current` is returned in.
```
# Note: For 2D simulations, multiply by device depth to get total current
device_depth = 50  # µm (assuming device extends 50 µm in z-direction)
I_ac = ac_current.sel(v=bias_voltage).values * device_depth
```

daquinteroflex:
Thanks @damiano-flex looks very good, just a few thoughts on a first pass

- Normally we add a warning of the estimated about of flexcredits at the top of the notebook, can you add this? Do you have an estimated time for our own internal reference?
- Is it worth reusing or slightly modifying a diagram from the charge notebook? Same setup?
- How do you normally do the thumbnail illustration @tomflexcompute ?
-  Is it worth also illustrating how the frequency domain result changes at different bias voltages / depletion amounts ie something like [ 2V, 4V...]? So basically multiple bias voltage lines in your bode plot? Maybe you can warn on the amount of flexcredits for that specific analysis? 
- Is it worth demonstrating why/ to what extent the small perturbation can change the frequency response? So like say +-0.1V perturbation around 1V might be interesting maybe?

Out of curiosity can't remember if we can also support electroabsorption as well so say around +0.7V rather than just depletion?

momchil-flex:
> * Normally we add a warning of the estimated about of flexcredits at the top of the notebook, can you add this? Do you have an estimated time for our own internal reference?

We haven't been very consistent about this - and the original idea was to add the warning only if it's more than 1FC. And the charge solver is currently always charged at the minimum cost of 0.025FC.

tomflexcompute:
@damiano-flex please edit the corresponding rst file (`docs/case_studies/pic_active.rst` I assume?) and notebook metadata (following instructions [here](https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e)). 

Could you prepare a schematic thumbnail for the notebook? I can prepare one for you if you prefer.

marc-flex (commented):
Thanks @damiano-flex for preparing this. It looks great!

Some comments:

- You're saying that you're running a reverse bias because this creates a strong field and reduces dark current (in the BC section). Note that dark current **increases** with high electric field so I'm not sure that point is necessarily true.
- In physical interpretation you're saying that the intrinsic region is fully depleted. Maybe this is a bit ambiguous language (i.e., what do you mean completely depleted? No electrons/holes in this region?). 
- Also, in that same line, you're saying `W` is approximately equal to the intrinsic layer. What is `W`? Is it related an equation? If so, what equation? Also, if there's an equation maybe you could also compare what the result using that equation would be an compare
- I'd remove the visualization of the DC solution? It doesn't seem to add much value
- So far you have demonstrated the effect of frequency on conductance/capacitance.It'd also be interesting to compare capacitance/conductance at two different bias levels to check the effect of bias over those parameters.

damiano-flex:
- Added feature picture, and edited the metadata.
- Added theoretical capacitance.
- Improved comment structure and content.
- Removed DC solution.

damiano-flex:
I updated the notebook with the following changes:
- Modified comments structure and context.
- Removed DC visualization.
- Added a temporary thumbnail picture.
- Modified the metadata of the notebook.
- Added analytical capacitance.

tomflexcompute:
@damiano-flex what do you think about this image?

image

Btw the notebook title (h1 header) only capitalizes the first letter of the first word.

damiano-flex:
> @damiano-flex what do you think about this image?
> 
> image
> Btw the notebook title (h1 header) only capitalizes the first letter of the first word.

@tomflexcompute the picture is awesome! If you want I can directly plug it in.

P.s. I'll fix the title too.

tomflexcompute:
> > @damiano-flex what do you think about this image?
> > image
> > Btw the notebook title (h1 header) only capitalizes the first letter of the first word.
> 
> @tomflexcompute the picture is awesome! If you want I can directly plug it in.
> 
> P.s. I'll fix the title too.

Yes feel free to use this image in the notebook. I didn't check but assume the notebook metadata has been added? If not, please do so following [this](https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e) or I can add it for you if you prefer.

damiano-flex:
@tomflexcompute Yes, I already set up the metadata. Titles are okay and the picture is changed.

marc-flex:
> @damiano-flex what do you think about this image?
> image

Hi @tomflexcompute, I think it is meant to say "intrinsic layer" instead of "depletion region"

tomflexcompute:
> > @damiano-flex what do you think about this image?
> > image
> 
> Hi @tomflexcompute, I think it is meant to say "intrinsic layer" instead of "depletion region"

Oh I see. See the updated version here 
image

damiano-flex:
Picture updated

marc-flex:
Thanks @damiano-flex this is looking great!

Some minor comments:

- I'd be careful with the sentence "At high doping levels, carrier-carrier interactions cause the effective bandgap to shrink". Bandgap narrowing has many effects, being one the most relevant the appearance of intermediate bands which in effect reduce the bandgap. So instead I'd just say something like: "Since we have high doping levels bandgap narrowing is expected to impact the solution. We can take this into consideration with the model..." or something along those lines
- Still the dark current comment. The larger the reverse bias, the larger the electric field and the larger the dark current. I think it'd be best to just remove that bullet point
- Also, I'm not sure about the bullet point "Minimize junction capacitance". The theoretical capacitance doesn't depend on the applied bias so where is this coming from?
- For the calculation of the theoretical capacitance you can use `td.EPSILON_0` instead of defining a new `eps0`. You could also use `eps_si = Si_medium.charge.permittivity`. That way it doesn't feel like we're performing a trick 
- A difference of 2% is awesome. Did this get better with mesh refinements? Also, if we increase the width by the Gaussian doping width (`junction_width` in the notebook) does it match better?
- In the conclusion list, I'd remove the first point or phrase it differently. More than designing what we're doing is analyzing a component. 
- I'm also not sure about including "photodiode" everywhere. What I mean is: we aren't demonstrating anywhere the properties of the diode with light. It's all in the dark. So if someone were to go through the example thinking they'd get some of characteristics under certain lighting conditions they'd be disappointed. Maybe @momchil-flex or @tomflexcompute have better criteria

damiano-flex:
1) Agreed. Modified accordingly.
2) Agreed. Bullet point removed.
3) As far as I understand, the junction capacitance in principle depends on applied bias, as increasing the reverse bias increases the width of the depletion region. Since Capacitance $C \approx \epsilon A / W$ (where $W$ is the depletion width), as Voltage increases $\rightarrow$ Width increases $\rightarrow$ Capacitance decreases. Maybe SSAC linearization doesn't catch this?
4) True agreed.
5) No, it doesn't get better with mesh refinement. It might be because the capacitance formula is an approximation, and I am wondering if it is expected to approach that value. Maybe by using constant doping with a strong reverse bias? (this way the depletion region $W$ is very straight).
6) Seems reasonable. Changed it to "**Defined the geometry of a silicon PIN photodiode** with realistic doping profiles and material parameters."
7) I understand, we could go for something like "Electrical Characterization of a PIN Photodiode".

damiano-flex:
Regarding 7, we are actually specifying in the title that we are doing a small-signal AC analysis of a PIN photodiode. Therefore, we are not claiming that we are doing any optical characterization, so this seems legit to me. This said, I am open to changes.

marc-flex:
>     5. No, it doesn't get better with mesh refinement. It might be because the capacitance formula is an approximation, and I am wondering if it is expected to approach that value. Maybe by using constant doping with a strong reverse bias? (this way the depletion region 

I was checking and the mesh is already very fine. As mentioned in a comment above; I'd try considering the gaussian wifth. I would add the length at which the doping falls by 2 or 3 orders of magnitude.

>     6. Seems reasonable. Changed it to "**Defined the geometry of a silicon PIN photodiode** with realistic doping profiles and material parameters."

Sounds good to me :+1:

PR #393, opened 2025-10-21
Alec/photothermal

alec-flexcompute:


FilipeFcp (approved):
Hi Alec, that’s a pretty cool notebook.
Everything is very clear; I just have a few minor comments.

It seems the image wasn’t uploaded.

Last line of the introduction: “siginal waveguide's phase change” → “signal waveguide's phase change”

Below cell 21: “the the control” → “the control”

alec-flexcompute:
Everything should be addressed, thanks for the feedback and help!

tomflexcompute (approved):
Thanks @alec-flexcompute for the great addition. 

It seems like the notebook doesn't involve any mode solving so `Mode analysis` probably doesn't need to be included in the features list?

alec-flexcompute:
Good catch! Fixed.

PR #395, opened 2025-10-23
Updating grcwa phase in multi-level grating example

momchil-flex:
In 2.9 we changed the diffraction amplitudes definition to have a phase origin at the center of the monitor, rather than at the corner of the simulation domain.


tomflexcompute:
In the second plot the comparison seems quite poor?

momchil-flex (commented):
Ah, that is exactly what I was fixing but somehow the updated plot wasn't committed (now after rerunning the notebook, there's also irrelevant updates, but that's how notebooks are...)

The agreement is still not perfect away from the 0 and +/-1 order, but I think it's acceptable. This is what it looks like in the currently published version before we changed the phase definition

image

tomflexcompute (approved):
Thanks. What's the underlying issue for not having a better match?

momchil-flex:
I am actually not sure. I wanted to study this a bit but when I doubled `space_above` and `space_below`, the match between the phase results became bad again. It seems that for some reason it works for the preset spacings, but maybe the reason why it's not good at high orders is related to why it breaks when the spacing is increased.

I'm hoping @weiliangjin2021 could provide some insight (Shashwat made the original notebook). I see that the position of the monitor seems to be accounted for

```
# position above the simulation domain's bottom where we're sampling
z_offset = monitor_t.center[2] - (-sim_size[2] / 2)
amps_grcwa_xy, _ = obj.Solve_FieldFourier(layer, z_offset)

# Extract grcwa results at orders corresponding to those computed by Tidy3D
amps_grcwa_xy = [
    np.array(rcwa_order_index(data_t.orders_x, data_t.orders_y, obj, amps)[1])
    for amps in amps_grcwa_xy
]
```

But I don't see the position of the source entering grcwa in any way.

PR #396, opened 2025-10-23
Upload TFLN EOM Tidy3D notebook

FilipeFcp:
Hi all,

This is the [PF example](https://docs.flexcompute.com/projects/photonforge/en/latest/examples/EO-MZM.html) adapted to Tidy3D.


yuanshen-flexcompute (commented):
Nice! Here are my comments:
* In preamble (first cell), the link for "mode simulation" currently links to API page. I think maybe it should link to our tutorial, like (https://www.flexcompute.com/tidy3d/examples/notebooks/ModeSolver/)?
* Cell [2] and onwards: since the code cells are quite large, I suggest to include additional lines of comments to describe what each logical grouping of code is doing. For example "Defining structures", "Defining grid specification" etc. The comment doesn't have to be that detailed, but I think it helps the reader parse the large code cells a bit better.
* Cell [3]: I suggest modifying last line to `_=mode_solver_opt.plot()` to suppress the Jupyter cell printing the matplotlib axes.
* Cell [5]: Maybe add a little more comment on the mode solver result? Like what are the important quantities in the result that motivated the mode solver.
* Cell [6]: I usually define `LossyMetalMedium` like `metal=td.LossyMetalMedium(conductivity=41, frequency_range=(f_min, f_max))` where `f_min, f_max` is the RF frequency range. I usually leave the `fit_param` untouched (default).
* Cell [7]: For the overall RF grid spec, you can use `min_steps_per_sim_size` when the simulation is sub-wavelength to better control the grid size (instead of `min_steps_per_wvl`). I usually start with values between 30-100. 
* Cell [14]: Perhaps a bit more explanation about what each block of code is doing? In particular, I think readers might be interested in the purpose of the `MeshOverrideStructure` and why the mesh needs to be so dense
* Since you re-calculate the CPW mode in the V-pi L section, is it necessary to calculate the original CPW mode without MZM in the first place? I think you could probably skip straight to the CPW mode with MZM. For basic CPW mode calculation, you could probably just refer readers to the separate tutorial (https://www.flexcompute.com/tidy3d/examples/notebooks/CPWRFPhotonics1/). 
* Cell [19]: There is a commented-out line that could probably be cleaned up
* Cell [19]: Again, could probably add `_=` to the final line to prevent printing of matplotlib object
* Cell [19]: If I understand correctly, the plot is supposed to show the shifted RF field? If so, I suggest to zoom into the region around y=0, and overlay the optical waveguide structure into the plot to make the overlap calculation clearer.
* For the calculation using the Pockels effect, could we include relevant equation(s) in the text to make the calculation clearer to the reader?
* Cell [20]: Again, I suggest to zoom into the y=0 region overlap with optical waveguide.
* Cell [21]: I am not entirely clear on the purpose of `eps_o_array` based on the comment
* Cell [21]: Could we also make it clearer to the reader that we are calculating n_eff as a function of applied voltage, with the assumption that RF field strength scales with voltage? I am not sure that it is explicitly mentioned anywhere

Thanks for this notebook!

FilipeFcp:
Thanks @yuanshen-flexcompute for the careful revision.
I have addressed all comments

FilipeFcp:
@tomflexcompute
I adapted the model to match the paper and also added some other results. I think the matching is pretty reasonable now.

tomflexcompute (approved):
Great @FilipeFcp !

PR #397, opened 2025-10-28
uploading far-field notebook with non-deacying field

FilipeFcp:
@momchil-flex, @tomflexcompute, and I were discussing last week the limitations of far-field projection for this scattering structure, where it’s not possible to use a closed 3D monitor and the fields don’t decay at the boundaries.

While exploring this, we noticed there’s a compromise between the simulation size and the `window_size` argument, so we decided to publish a quick tutorial about it. Any insights or comments you might have are very welcome :)


momchil-flex:
Any idea why you're getting this? I can't reproduce.

image

momchil-flex:
This is pretty interesting! Could you add a plot of the simulation setup so it's easier to understand?

I am surprised that the 3D projection monitor does poorly. Could you check what happens if you make it just a little larger than the scatterer in xy, rather than spanning almost the entire simulation domain?

In the last plot, you might want to play around with the figsize, maybe decrease the horizontal size because the font is quite small? And that should create more vertical space to avoid this overlap.

image

FilipeFcp:
I did some tests with the 3D monitors of 1.2 µm and 2 µm side (the scatterer has 1 µm side). The results seem worse.
Note that I need to exclude the negative z surface.

image

momchil-flex:
> I did some tests with the 3D monitors of 1.2 µm and 2 µm side (the scatterer has 1 µm side). The results seem worse. Note that I need to exclude the negative z surface.

Ok yeah I think I understand why this is the case, fundamentally the fields don't decay which is always going to be a problem.

tomflexcompute (approved):
Thanks @FilipeFcp this is very helpful to clear up some confusion from users. Looks ready to go.

1. I think the correct rst file should be `docs/features/data_visualisation.rst`.
2. The section title `Simulation setup` -> `Simulation Setup`

PR #401, opened 2025-11-07
uploading APD example

FilipeFcp:
Hi all,

I’ve added some text and the image.
The mode simulation looks good, I ran a few tests, and the results seem quite solid. It’s a bit different from what’s shown in the paper, but if I understood correctly, they simulate the coupling from the Si waveguide into the Ge region. Nevertheless, the results for the photocurrent appear consistent. 

The main discrepancy occurs around 11 V, which seems to correspond to the breakdown voltage. But I couldn't figure out why.


tomflexcompute (commented):
Thanks @FilipeFcp and @marc-flex on making this possible. A few comments:
1. "Photodiode" on the title -> "photodiode" 
2. In the beginning, add a brief overall introduction about avalanche photodiode
3. This is more for @marc-flex , can we add some convenient functions to calculate the optical absorption to reduce the codes in cell[12]?
4. Explain a little bit on the physics of dark and bright currents as well as what the final result implies.

marc-flex (approved):
Thanks @FilipeFcp for turning my draft into this nice notebook! I left some minor comments (most of which are my fault to begin with). It'd also be good if @momchil-flex could have a quick look

- In cell 2 I left a T0 = 280. I think it is safe to just use 300. Either way we should probably only leave 1 to avoid confusion.
- In cell 3 maybe remove comments (e.g., `alpha_n_inf=7.03e5/8000,`) that I used to try things out. Actually, I left a lot of comments scattered throughout the notebook. So sorry about it
- in cell 20 I'd change the comment to say that we're only visualizing the x component of the electric current. As an alternative we could compute the modulus of the current density vector and plot that instead?
- In cell 21 maybe add label for the x axis? i.e., applied bias

marc-flex:
> This is more for @marc-flex , can we add some convenient functions to calculate the optical absorption to reduce the codes in cell[12]?

We should definitely think of something, especially as we add more other effects such as the TPA that @momchil-flex has been working on.

FilipeFcp:
Thanks guys, I think it is adressed

FilipeFcp:
I just noticed that chanfing T0 to 300 makes the result different:
image

I will change it to 280

momchil-flex:
Thanks @marc-flex and @FilipeFcp !

> * In cell 3 maybe remove comments (e.g., `alpha_n_inf=7.03e5/8000,`) that I used to try things out. Actually, I left a lot of comments scattered throughout the notebook. So sorry about it

Indeed, these comments are still there. Best to remove, but how can we justify the values that are actually used?

image

Probably good to mention what some of the symbols are, like epsilon'', q, and maybe use hbar omega instead of h nu given that the abs power equation already uses omega.

I notice that the definition of `P_in` is buried in code only. Maybe worth pointing out somewhere in the text that the results depend on the optical power, which we set to so and so watts.

Otherwise, looks great! I think indeed we should introduce some convenience for the absorption, but this PR can go without this.

tomflexcompute (approved):

FilipeFcp:
Thanks, guys. I think it is all addressed

PR #402, opened 2025-11-14
Uploading diffraction beam splitter example

FilipeFcp:


FilipeFcp:
Hi Guys,

I created an ID notebook based on some interesting results I found while working on a customer model. I thought this might be a good candidate for the Autograd example library.

Could you please take a look when you have the time? Your comments on whether this is a good fit for the library are greatly appreciated.

If you decide it's not quite right for the public examples, no problem, we can just keep it as an internal reference.

tomflexcompute (commented):
Thanks @FilipeFcp this is a very informative and practical example that we should definitely include in the documentation. A few comments:

1. For the figure of merit, can you explain the exact function you defined and why you chose this form in the markdown comment above it?

2. For the diffraction power plot, I would suggest a higher contrast colormap like `hot` for example. `viridis` is little weak in my opinion.

3. The final design performs very well, but it seems to contain small features. Maybe add a comment about it, or even better, use something to characterize (qualitatively maybe) the minimal feature size. Then put a note that to ensure minimal feature size is compliant with fab, users can play with the fab panelty, filter, etc.

docs/features/autograd.rst
yaugenst-flex, line 1: Docs entry has inconsistent formatting, none of the other entries include the `.ipynb` extension

yaugenst-flex (commented):
Thanks @FilipeFcp this is really cool! Couple of comments:

- In the optimization output, there is `penalty_weight = ...`, but this isn't defined anywhere. I assume this is from some prior runs?
- Cell 18 markdown: The `DiffractionMonitor` link is broken (no URL defined?)
- The monitors `mnt_out` and `mnt_side` are unused - remove?
- `post_process` imports `matplotlib` and `numpy`, but these are already imported at the top of the notebook so should be removed.
- In the final analysis you use `make_filter_and_project` with a radius of 0.5, but in Cell 10 you use 0.2 - is this intentional?
- I'm a bit confused about the coordinate span of the custom medium. The permittivity data only seems to cover `(length/2 - pixel_size/2) ~ 2.5um` but the box geometry is `(10, 10)` - is this intentional? Even if it is it seems a bit confusing.
- Typo: `optmization` -> `optimization` in cell 7 markdown
- Task name is `phase_mask_example` which I guess is leftover from the notebook this was adapted from?

FilipeFcp:
Thanks, everyone. I think it's all addressed.
@yaugenst-flex, the extra size for the geometry is to avoid spamming warning messages due to the boundaries close to the PML.
I added a comment about it

tomflexcompute (approved):
Thanks @FilipeFcp . Looks great!

PR #403, opened 2025-11-17
Adding other TCAD data to ChargeSolver.ipynb

marc-flex:
Adding other TCAD data to our existing ChargeSolver.ipynb notebook


FilipeFcp (approved):
Thanks Marc!
Users have been asking for this, so it will be very useful.

momchil-flex (approved):
Cool! While you're at it, could you update to avoid the deprecation warnings?

image

I also wonder what this is about:

```
module 'tidy3d.config' has no attribute 'get_manager'
```

do you know @yaugenst-flex ?

yaugenst-flex:
> I also wonder what this is about:
> 
> ```
> module 'tidy3d.config' has no attribute 'get_manager'
> ```
> 
> do you know @yaugenst-flex ?

Mysterious, no idea really.. the only place where that is imported is in [tidy3d/web/cli/app.py](https://github.com/flexcompute/tidy3d/blob/2477d968ead50b763c49c31e509cf8e48b3803f5/tidy3d/web/cli/app.py#L15) but I've never seen this not work. Very weird.

yaugenst-flex:
So when I run this from `develop` I get: 
image

This suggests to me that the error might be coming from `tidy3d-extras`? I don't think there is anything in Python client that would cause this warning

PR #405, opened 2025-11-20
refactor(tidy3d): FXC-3960 Update notebooks to use new config API

marcorudolphflex:
Changed to the new config API which is here basically only replacing the` td.config.logging_level` to `td.config.logging.level` in almost every notebook (and replacing `set_logging_level` in one case).


LedLEECalculation.ipynb
yaugenst-flex, line 1: why does this have such a crazy number of changed liens?

yaugenst-flex (commented):
Thanks @marcorudolphflex this is great! Only thing i'm wondering is why the diff for some of these notebooks is `+1 -1` (as expected) and for others it's huge?

Also, could you have a look at the spell check and maybe fix those or add them to the dictionary? `spellcheck.py` has an interactive mode that makes this pretty easy.

LedLEECalculation.ipynb
marcorudolphflex: PyCharm doing PyCharm things.. now it should be clean

marcorudolphflex: and will do the spellcheck...

yaugenst-flex (approved):

PR #408, opened 2025-11-20
feat(tidy3d): FXC-3718-docs-and-examples-for-container-aware-run

daquinteroflex:
Sorry we should have made clearer how the notebooks get synced to the webcenter.


FilipeFcp:
@marcorudolphflex, what is the process for canceling a simulation batch, or reloading it in case of disconnection, using web.run?
Is there something similar to Batch.load, and Bactch.delete?

marcorudolphflex:
> @marcorudolphflex, what is the process for canceling a simulation batch, or reloading it in case of disconnection, using web.run? Is there something similar to Batch.load, and Bactch.delete?

There is no opportunity to do such things with web.run as it should be just the easy/convenient way to directly gather the data...
We could think about allowing such (nested) containers in Batch as well. What do you think, @yaugenst-flex?

FilipeFcp:
From time to time, a user has a situation where they need to cancel a batch simulation or lose their current Python session (due to a disconnection or a kernel crash) and need to reload the simulations.

There is also the issue of loading the simulation after running it. The Apodized notebook has more them 100 tasks. If one wants to look at the data again, that would be very difficult.

In my opinion, if it were possible to save a file similar to batch.hdf5, containing the task IDs, could potentially save some headaches.

marcorudolphflex:
> From time to time, a user has a situation where they need to cancel a batch simulation or lose their current Python session (due to a disconnection or a kernel crash) and need to reload the simulations.
> 
> There is also the issue of loading the simulation after running it. The Apodized notebook has more them 100 tasks. If one wants to look at the data again, that would be very difficult.
> 
> In my opinion, if it were possible to save a file similar to batch.hdf5, containing the task IDs, could potentially save some headaches.

Ah, that's what you mean. 
Internally, web.run does make a batch (if more than one unique simulation) with the simulation hashes as task names and exports a batch.hdf5. You could use it like:
```
    web.run({"sim1": sim, "sim2": sim2}, path=out_path)

    batch = td.web.api.container.Batch.from_file(f"{out_path}/batch.hdf5")
    sim2job = batch.jobs[sim2._hash_self()]
    results2 = sim2job.load(path="sim2.hdf5")
```
This could be surely more user friendly (regarding _hash_self...)

marcorudolphflex (approved):

FilipeFcp (approved):

FilipeFcp:
Thanks @marcorudolphflex, that is good!

PR #409, opened 2025-11-20
docs: minor notes about syncs and releases

daquinteroflex:
apologies again


FilipeFcp (approved):
Thanks! It is pretty clear now.
And there is no need to apologise :)

I think there is a typo here: - The `develop` branch is syced .. -> The `develop` branch is synced ..

PR #413, opened 2025-11-27
added Autograd25WaveguideCrossing.ipynb

mahlau-flex:
Added a notebook showcasing the new smoothed projection (see paper here: https://arxiv.org/pdf/2503.20189). This method is exempified with a waveguide crossing.


mahlau-flex:
resolved spelling mistake

yaugenst-flex (commented):
Thanks @mahlau-flex this looks great overall! Just a few notes:

- Could you add this to the notebook index [here](https://github.com/flexcompute/tidy3d-notebooks/blob/develop/docs/features/autograd.rst)
- The initial learning rate of 1 seems really high and you can also see the objective function dipping in the first iteration due to that. Have you tried with a lower learning rate?
- I think we should more prominently cite the paper that the parametrization is from at the top.

yaugenst-flex:
@tomflexcompute @FilipeFcp could one of you have a look too?

mahlau-flex:
@yaugenst-flex I have incorporated your previous comments into the notebooks

FilipeFcp:
Thanks @mahlau-flex, this is a very nice example, very didactic!

I have just a few comments:

- There is an image in the metadata, but it is not uploaded. It would also be good to insert it in the first markdown cell to stay consistent with the other examples.
- Is it possible to remove the warning after the import cells?
- It would be nice to include the legends in the final plot (I believe `plt.legend()` is enough).

- That is personal, but I found the explanation before cell [7] a bit confusing. I think it could be more direct. 

Very small nitpicks:

- Docstring in cell 5: “as function” → “as a function”.
- Before cell [14]: “initialize adam optimizer” → “initialize Adam optimizer”.
- Cell [24]: “initial_waveguide_crossing” → “final_waveguide_crossing”.

- This warning is a nice addition.

  > Important note: we use `autograd.numpy` instead of regular `numpy`; this allows `numpy` functions to be differentiable. If one forgets this step, the error may be a bit opaque — just a heads up.

  I have assisted a few users who forgot this step. Maybe we should consider adding it to other ID notebooks as well.

On a separate note, @yaugenst-flex, can we remove these notes from the other ID notebooks?

“Note: native autograd support is an experimental feature in tidy3d 2.7. To see the original implementation of this notebook using jax and the adjoint plugin, refer to this notebook.”

mahlau-flex:
@FilipeFcp thanks for the feedback! I have incorporated your points into the notebook

tomflexcompute (approved):
Very nice @mahlau-flex ! I didn't check the notebook metadata but please make sure it has been added by following the instructions [here](https://www.notion.so/flexcompute/How-to-add-metadata-to-a-case-study-or-tutorial-notebook-696f9dd642f14964af4eb1242226451e) if you haven't.

mahlau-flex:
Hi @tomflexcompute, yes i added the correct metadata information

PR #414, opened 2025-11-28
fix: removing warnings in Autograd notebook 17

momchil-flex:
Also updated a deprecation warning in the optimizer in the end.

The optimization result is slightly better than previously.


yaugenst-flex (approved):
Nice thanks!

PR #417, opened 2025-12-08
uploading hexagonal band diagram calculation example

FilipeFcp:
Hi all,
I’m uploading the example for the hexagonal band diagram calculations.
There is one cell with some theory that was entirely generated by AI. I’m still debating whether we should keep it or remove it to make the example shorter.

@momchil-flex, feel free to take a look if you have time


momchil-flex:
Thanks @FilipeFcp !

Just a couple of things:
- There seems to be a missing link 
image

- Maybe mention in one of the markdown cells around the simulation creation / plotting that 5 extra monitors are automatically added
- The part of the code that computes the spectrum from the FFT, and e.g. the FTs variable is unused
- My biggest question: MPB is pretty good and fast for modes below the light line. I think one of the main advantages of doing this in FDTD is to be able to also get the modes above it. Have you not been able to get nice results for the quasi-guided modes?

alec-flexcompute (commented):
Great example again @Filipe! A couple things:

1. I think Callable is imported but not used

2. In the second cell, you import numpy a second time

3. the **args argument in the ftField function isn't used - maybe you can drop?

4. When all is edited remember run all cells in order so the cells are ordered

5. Momchil brings up a good point with modes above the light line. It could be interesting to see, but I'm not sure how much time it would take to get good results

tomflexcompute (approved):
Great addition to answer a pretty common question! Thanks @FilipeFcp . The notebook is also well organized with a great teaching style.

The header "K-points definition" -> "K-points Definition".

momchil-flex:
Nice. Just one thing, I would word "source decay time and mask the data immediately after the decay" a bit differently since to me this sounds like you *remove* the data after the source decay. Maybe "source decay time and only use the data after the decay."

And it still seems to me that `FTs` is completely unused so the relevant code can be removed?

FilipeFcp:
Thanks @momchil-flex!

I was plotting the FT for debugging, and forgot to remove it in the final version

PR #429, opened 2026-01-05
metadata fix

FilipeFcp:
I noticed some weird things in the example library:

image

It seems it is just the metadata of this notebook that the feature entry was 'Charge' instead of ['Charge']. It is a simple fix so I just did it


tomflexcompute (approved):
Great catch!

PR #437, opened 2026-01-16
fix params history

FilipeFcp:
Hi Tom,

A user noticed in [this](https://www.flexcompute.com/tidy3d/examples/notebooks/Autograd9WDM) notebook that all the entries in `params_history` are equal. This happens because they are all (except the first) references to the same object.

I noticed this also happens in other notebooks, so I replaced occurrences of `params_history.append(params)` with `params_history.append(params.copy())`.


tomflexcompute (approved):
Great catch thanks @FilipeFcp !

PR #440, opened 2026-01-27
Update Segmented CPW notebook with LXT collaboration data

yuanshen-flexcompute:
Updated the segmented CPW notebook to include comparison with experimental data from LXT.


tomflexcompute (approved):

tomflexcompute (commented):
Actually @yuanshen-flexcompute you need to edit the mapping json file in the misc folder too.

PR #441, opened 2026-01-28
upload Silicon Hybrid example

FilipeFcp:
This is an example that I explored for the optical meeting.
Unfortunately it is pretty hard to compare exactly with paper results, because many parameters are not provided.


alec-flexcompute (approved):
This is an awesome notebook @FilipeFcp ! Excellent organization and flow (I also liked the summary of workflow at the top). Just a few variables that are defined but I don't think are used:

Importing pandas
fwidth
clad_gap
sio2_metal_gap
pad_x, pad_y, pad_z

FilipeFcp:
Good catch, thanks @alec-flexcompute !

tomflexcompute (approved):
Very nice addition @FilipeFcp . All looks great. Could you get rid of the warning in the tidy3d import by removing the legacy config file?

PR #443, opened 2026-01-29
demo: precision in Simulation parameters

daquinteroflex:


momchil-flex (approved):
I would word this even stronger in the direction that this should very rarely be needed. Something like, the hybrid precision is practically sufficient in almost all cases, double precision should only be explored in large simulations when very high accuracy is desired, and even then it might not have a noticeable effect over other sources of numerical error like the finite grid.

daquinteroflex:
Modified it to follow your wording as suggested.

PR #445, opened 2026-02-02
Simple RF parameter sweep demo

yuanshen-flexcompute:
Simple demo notebook to show how to run a parameter sweep for RF. Even though the workflow is largely similar to Tidy3D, RF users might benefit from a RF-focused example for them to follow.


alec-flexcompute (approved):
Excellent notebook @yuanshen-flexcompute ! I have no criticisms. Maybe a mild suggestion would be to add a bit of a description about the simulation itself. But since that's beside the point and not what users would be looking for, it's totally fine if you don't add one.

yuanshen-flexcompute:
@tomflexcompute the auto-check is failing for some reason. I made sure to run `ruff format` and `ruff check --fix`. Do you have any ideas?

tomflexcompute (approved):
Thanks @yuanshen-flexcompute . Looks great.

As discovered by Greg, I think it's just a ruff version issue?

PR #447, opened 2026-02-13
Momchil/heat warnings

momchil-flex:
Updating HeatChargeSimulation notebooks to avoid some recently introduced warnigns.

@marc-flex I'll also rerun the notebooks to check if any outputs have changed, but at the current state it will be cleaner for you to see the differences.

Ignore the extra config warning and the big components printout in the thermally tuned heater notebook, this was a remnant from my testing.


HeatSolver.ipynb
marc-flex, line 1: Don't we prefer to have `unstructured=True` in cell 14 since structured will be deprecated?

PlasmonicPhotothermalHeating.ipynb
marc-flex, line 1: Same thing for this one

ThermallyTunedRingResonator.ipynb
marc-flex, line 1: Maybe remove the configuration warning? 

Also, we have all that output when importing tidy3d

marc-flex, line 1: Also in this one we have structured temperature monitor which we may as well make unstructured.

marc-flex (approved):

ThermallyTunedRingResonator.ipynb
momchil-flex: Yeah in my PR description I mentioned you can ignore these, just needs a rerun.

momchil-flex:
Yeah so I actually got a bit confused whether `unstructured` will be deprecated or whether it will just not be default anymore? The warning just says that the default will switch. We're deprecating `conformal`, but are we deprecating `unstructured` too?

In the current changes, the notebooks just preserve the existing monitors: wherever `unstructured` was not explicitly set, it was changed to the explicit current default `False`.

marc-flex:
> but are we deprecating `unstructured` too?

I think @dbochkov-flexcompute can confirm but I think we'll end up removing `unstructured` altogether since there wasn't any good reason to keep it.

dbochkov-flexcompute:
> > but are we deprecating `unstructured` too?
> 
> I think @dbochkov-flexcompute can confirm but I think we'll end up removing `unstructured` altogether since there wasn't any good reason to keep it.

Yeah, I think it would make sense to deprecate it as well in the favor of unstructured=True always, could be a good to time to do it together with conformal. Initially it was unstructured=False always because we didn't have infrastructure to handle unstructured data. Returning uninterpolated data in the original form is always preferred I believe, moreover this interpolation contains some heuristic automations that is difficult control as it happens during solver run. If data on a Cartesian grid is needed the user can do that manually after they get results back

momchil-flex:
Is there some argument for leaving unstructured=False as a way to get a "compressed" version of the data? There's no other way to down-sample currently. @dbochkov-flexcompute @marc-flex ?

marc-flex:
> Is there some argument for leaving unstructured=False as a way to get a "compressed" version of the data? There's no other way to down-sample currently. @dbochkov-flexcompute @marc-flex ?

You mean defining the monitor with `unstructured=False` would generate output to a (coarse) cartesian grid? If so, I'm not opposed except I would maybe add some warning? Or maybe simply update the description?

momchil-flex:
Oh. Actually the unstructured data is significantly smaller at least in these two simulations from the HeatSolver notebook I had a look at! I guess the structured one interpolates on a pretty fine grid.

```
Data size comparison (unstructured vs structured):

Simulation                Unstructured (KB)    Structured (KB)      Ratio     
---------------------------------------------------------------------------
planar                    10.84                105.19               9.70      
cylinder                  1546.64              86977.73             56.24   
```

Sounds like we can remove the option then and provide proper downsampling some day if we want that.

marc-flex:
Yes, if I remember correctly I think it takes the smaller cell size to create the cartesian grid

dbochkov-flexcompute:
> Oh. Actually the unstructured data is significantly smaller at least in these two simulations from the HeatSolver notebook I had a look at! I guess the structured one interpolates on a pretty fine grid.
> 
> ```
> Data size comparison (unstructured vs structured):
> 
> Simulation                Unstructured (KB)    Structured (KB)      Ratio     
> ---------------------------------------------------------------------------
> planar                    10.84                105.19               9.70      
> cylinder                  1546.64              86977.73             56.24   
> ```
> 
> Sounds like we can remove the option then and provide proper downsampling some day if we want that.

Yeah, I guess this is basically adaptive vs Cartesian grid. One thing is that Cartesian grid is definitely much easier to downsample if needed, but it is also not impossible for unstructured data

momchil-flex:
> Yeah, I guess this is basically adaptive vs Cartesian grid. One thing is that Cartesian grid is definitely much easier to downsample if needed, but it is also not impossible for unstructured data

Indeed, apart from easier, it is *faster*. I would say, we can switch the default, but there's no rush to remove the structured option. 

The last cell in the HeatSolver notebook used to take 2s with structured monitors. With unstructured, it took 2m20s, and that was after [this improvement](https://github.com/flexcompute/tidy3d/pull/3299). Before that it was over 4 minutes. I experimented a little bit in improving further, but it's hard to avoid the vtk operations in the symmetry expansion and in `.sel` without significant refactor. The current state of the notebook in this PR was run with unstructured=True if you want to take a look.

I could revert to `unstructured=False`, or we just accept the significant slowdown?

dbochkov-flexcompute:
> Indeed, apart from easier, it is _faster_. I would say, we can switch the default, but there's no rush to remove the structured option.
> 
> The last cell in the HeatSolver notebook used to take 2s with structured monitors. With unstructured, it took 2m20s, and that was after [this improvement](https://github.com/flexcompute/tidy3d/pull/3299). Before that it was over 4 minutes. I experimented a little bit in improving further, but it's hard to avoid the vtk operations in the symmetry expansion and in `.sel` without significant refactor. The current state of the notebook in this PR was run with unstructured=True if you want to take a look.
> 
> I could revert to `unstructured=False`, or we just accept the significant slowdown?

The data at the finest resolution is pretty large after symmetry expansion (13.7M cells x 8 -> 109M cells). Might take some time to figure how to speed up operations in that case. For now, a quick workaround is to not do symmetry expansion in that cells, just calculate accuracy on reduced data, something like:
```
for ind, (_, sim_data) in enumerate(batch_data.items()):
    line_data = sim_data.data[0].temperature.sel(z=0, y=0)
    x = line_data.coords["x"]
    x_exact_sphere, temperature_exact_sphere = get_analytical(3, x)
    temperature_numerical = line_data.sel(x=x_exact_sphere, method="nearest").data.ravel()
    error = np.abs(temperature_numerical - temperature_exact_sphere)
    errors.append(np.max(error))
    ax[0].scatter(x_exact_sphere, error, label=f"dl = {dl_refine[ind]}")
```

momchil-flex:
@FilipeFcp can you take a look at the CPO notebook (and any other ones if you want to)?

For now I had to leave the unstructured mesh as it was originally set. However I made the y-domain for the first two simulations much smaller as we discussed (just size `dl`, note that even a 2D heat simulation is 3D in the solver so setting it as 0 wouldn't be too different). I also removed your log level setting to ERROR to make sure I can actually clear as many warnings as possible (for example, setting the size in the other two dims to 0.99 times the geometry span).

I also reduced the number of time steps from 1000 to 100 because rerunning this was in many cases causing problems either on the server (way too much data needs to be written to disk) or in my local env (OOM in plotting, slicing, etc.)

FilipeFcp (approved):
Thanks @momchil-flex . It seems good!

PR #451, opened 2026-02-20
Radiative loss update

momchil-flex:
- Remove usage of deprecated mode web api
- Unify some sequential calls in to Batch calls
- Only store E-fields so only half the data is downloaded


FilipeFcp (approved):
Thanks @momchil-flex

PR #457, opened 2026-03-13
FXC-6319 notebook for source gradient introduction

marcorudolphflex:


tylerflex (approved):
This looks great! I only have two suggestions:

1. It would be helpful to add a schematic / diagram showing what is happening. For example just showing two thin simulation boxes, separated by a region of empty space with "field projection" followed by another "field projection" label for a focal point. I think it would help someone scanning this over to understand exactly what is going on.

2. We tend to include a few more links in the notebook intro cell to some other notebooks or relevant info, so I would perhaps suggest adding a few references to eg. an inverse design intro, and the other metalens inverse design examples.

groberts-flex (commented):
Thanks @marcorudolphflex, overall a great demonstration of the source gradient feature! A few comments on the notebook, mostly on polish.

1. The introduction might benefit from starting a little more generally to motivate the concept and utility of having source gradients. This could be done even in the scope of just two-level metasurfaces to motivate the notebook a little more generally before jumping into the details. You could mention some of the possible design goals for two-level metasurfaces (like aberration correction for example) and then talk about the typical ways two-level design is done without the source gradient like, for example, optimizing each metasurface individually for different target functions without the option to evolve them together.
2. The plot to show the meta surface setup is great. It might be good to roll that plotting code into a function to section it off from the rest of the notebook a little. And then the reader can ignore a bit the details of that while getting an idea for the notebook.
3. I believe stylistically (if I remember correctly), the section titles in the notebook capitalize each word. So "Objective function" would instead be "Objective Function".  The notebook title itself should only capitalize the first word, but that's how you have it already.
4.  Some of the text in between the code has some extra newlines in it. For example, this one probably doesn't need the newline between the two sentences.
"We use circular Si pillars on an SiO2 substrate.
Symmetry allows us to model only one quadrant of each metasurface in the simulation."

5. For the final projected axial power plot, is the z-axis lined up with the geometry plot earlier in the notebook. If so, are we expecting the focus to be at z=6um? It looks shifted up a bit in z.

6. For some of the prints, it might be nice to reduce the number of decimals for display purposes. For metrics like power, are they normalized to input power? For example, in the below print, is that 24% focused power?
"final absolute power = 2.3908e-01"

marcorudolphflex:
> Thanks @marcorudolphflex, overall a great demonstration of the source gradient feature! A few comments on the notebook, mostly on polish.
> 
> 1. The introduction might benefit from starting a little more generally to motivate the concept and utility of having source gradients. This could be done even in the scope of just two-level metasurfaces to motivate the notebook a little more generally before jumping into the details. You could mention some of the possible design goals for two-level metasurfaces (like aberration correction for example) and then talk about the typical ways two-level design is done without the source gradient like, for example, optimizing each metasurface individually for different target functions without the option to evolve them together.
> 2. The plot to show the meta surface setup is great. It might be good to roll that plotting code into a function to section it off from the rest of the notebook a little. And then the reader can ignore a bit the details of that while getting an idea for the notebook.
> 3. I believe stylistically (if I remember correctly), the section titles in the notebook capitalize each word. So "Objective function" would instead be "Objective Function".  The notebook title itself should only capitalize the first word, but that's how you have it already.
> 4. Some of the text in between the code has some extra newlines in it. For example, this one probably doesn't need the newline between the two sentences.
>    "We use circular Si pillars on an SiO2 substrate.
>    Symmetry allows us to model only one quadrant of each metasurface in the simulation."
> 5. For the final projected axial power plot, is the z-axis lined up with the geometry plot earlier in the notebook. If so, are we expecting the focus to be at z=6um? It looks shifted up a bit in z.
> 6. For some of the prints, it might be nice to reduce the number of decimals for display purposes. For metrics like power, are they normalized to input power? For example, in the below print, is that 24% focused power?
>    "final absolute power = 2.3908e-01"

Thanks for the feedback.
All of the comments should be addressed now.
Adjusted the field plot such that z=6 is the focus plane now (before it was relative from the field projection, not the lens), also added a line and legend that marks the focus plane in the plot.
Whenever a normalized value is given, I state them as "normalized" - otherwise just the absolute value is meant.

groberts-flex (approved):
thanks for the changes @marcorudolphflex, really cool notebook!

Only small thing I saw on a second pass is the spelling "focussed" to change to "focused".

The plot at the end does seem like the focus is being formed before the focal plane but this could be a geometry limitation of the whole system and the optimization is clearly working to improve power at the focal plane even if the focus is happening a little before it so I think it works great as a demonstration for how to set these things up.

Good to merge for me!

PR #458, opened 2026-03-16
refactor(notebooks): use internal adam and optimize helper - batch 1 (FXC-6310)

marcorudolphflex:
Replaces the optax adam with our adam and make usage of the optimize helper when easy possible. This is just the first batch of it with 4 out of 20 notebooks to make that reviewable.

Results did slightly change (proably due to solver changes?), but not decisive I would say.

Won't apply the optimize helper for more complex loops where the objective is dependent on the step_index (for now). We could think about extending the optimize function for such purpose. Not sure if we want to do this while not blowing the logic of this helper or if this is then really still convenient as you have to wire things less intuitive compared to plain loops?


yaugenst-flex (approved):
Nice!

tylerflex (approved):
thanks!

PR #460, opened 2026-03-17
refactor(notebooks): use internal adam and optimize helper - batch 2 (FXC-6310)

marcorudolphflex:
follow-up from https://github.com/flexcompute/tidy3d-notebooks/pull/458

Replaces the optax adam with our adam and make usage of the optimize helper when easy possible. This is the second batch of it with 5 out of ~20 notebooks to make that reviewable.

Results did slightly change (proably due to solver changes?).
Mentionable is this change in the objective for [Autograd10](https://github.com/flexcompute/tidy3d-notebooks/blob/develop/Autograd10YBranchLevelSet.ipynb) - will investigate if this is related to the optimizer which I don't expect.
Before:
image
After:
image

Won't apply the optimize helper for more complex loops where the objective is dependent on the step_index (for now). We could think about extending the optimize function for such purpose. Not sure if we want to do this while not blowing the logic of this helper or if this is then really still convenient as you have to wire things less intuitive compared to plain loops?


tylerflex (commented):
Thanks @marcorudolphflex also noticed a change in `‎Autograd15Antenna.ipynb‎`

Image

Image

the new one looks a bit better but it's a bit weird because the gradient norm looks the same between both but the magnitudes in the plots are different and th optimization curve is also different?

marcorudolphflex:
> Thanks @marcorudolphflex also noticed a change in `‎Autograd15Antenna.ipynb‎`
> 
> Image Image
> the new one looks a bit better but it's a bit weird because the gradient norm looks the same between both but the magnitudes in the plots are different and th optimization curve is also different?

Oh, good catch - I was more focussed on the objective here, which was even higher after the refactoring.
The gradient thing is super weird, I mean this comes even before the optimizer stuff (and my first change in this notebook). ̶ ̶J̶u̶s̶t̶ ̶r̶e̶-̶e̶v̶a̶l̶u̶a̶t̶e̶d̶ ̶i̶t̶ ̶o̶n̶ ̶d̶e̶v̶e̶l̶o̶p̶ ̶a̶n̶d̶ ̶g̶o̶t̶ ̶t̶h̶e̶ ̶s̶a̶m̶e̶ ̶a̶s̶ ̶i̶n̶ ̶t̶h̶i̶s̶ ̶d̶e̶v̶ ̶b̶r̶a̶n̶c̶h̶ ̶(̶h̶i̶g̶h̶ ̶g̶r̶a̶d̶ ̶n̶o̶r̶m̶)̶.̶ ̶T̶h̶e̶r̶e̶ ̶i̶s̶ ̶s̶o̶m̶e̶t̶h̶i̶n̶g̶ ̶I̶ ̶m̶i̶s̶s̶,̶ ̶l̶e̶t̶ ̶m̶e̶ ̶r̶e̶v̶i̶s̶i̶t̶ ̶t̶h̶a̶t̶.̶
I used the wrong branch for latter tests.

marcorudolphflex:
I could just confirm that the Antenna notebook changes are not related to the optimizer, but probably due to some recent changes in the gradients of `CustomPoleResidue`. I will revisit the numerical tests and see how the gradient norms look compared to the FD gradients.

marcorudolphflex:
The reason in this change is probably due to several changes and fixes in the autograd pipeline for medium we had in the last weeks.
Made a FD-comparison on the min grad value (max abs grad) and got
```
finite-difference check at most negative autograd entry:
	index = (83, 77, 0)
	params0[(83, 77, 0)] = 0.500000
	autograd grad = -8.262743e-04
	fd grad = -1.822359e-04
```
So the new gradient is much closer to this magnitude. Unfortunately, the norm is quite off.
I also reran `test_autograd_custom_pole_residue_numerical.py` and `test_autograd_medium_numerical.py`. The tests passed (regarding angle), but for dispersive mediums again the norms are partially off up to a factor of ~6, but there were also cases with a almost-matching norm. 
Guess we have to revisit that @groberts-flex.
Still I think we should merge this as this is not related to the optimizer.

groberts-flex:
After chatting with @marcorudolphflex a bit on slack about this, it sounds like the norm for these gradients isn't a regression and more something we haven't tested before. Overall, the gradients are much closer in norm than before so I will make a ticket to investigate this more closely. For now, it's great to see we are much closer in norm than before and the optimization performance is better!

yaugenst-flex (approved):
LGTM — optimizer swap is correct, good decisions on where to use optimize() vs manual loops, numerical diffs adequately investigated.

PR #461, opened 2026-03-19
updating Charge MMI example and PIC Components to use Photonforge layout

FilipeFcp:
Hi guys.

I tried using Claude code for converting the PIC notebook to photonforge. I think it worked pretty well.


tomflexcompute (approved):
Nice.

alec-flexcompute (approved):
Looks great! Not sure how important this is, but the first cell where imports are done is throwing the weird error, which I'm not sure what it's about

PR #463, opened 2026-03-25
refactor(notebooks): use internal adam and optimize helper - batch 3

marcorudolphflex:
follow-up from https://github.com/flexcompute/tidy3d-notebooks/pull/460

Replaces the optax adam with our adam and make usage of the optimize helper when easy possible. This is the third batch of it with 5 out of ~20 notebooks to make that reviewable.

results are roughly unchanged (often slightly better)


tylerflex (approved):
looks good, thanks!

PR #466, opened 2026-04-05
chore: rerun ThermoOpticDopedModulator with eps broadcast fix

daquinteroflex:
## Summary
- Rerun ThermoOpticDopedModulator.ipynb after fixing shape mismatch in `Simulation.epsilon_on_grid()`
- All 39 code cells execute successfully
- Depends on flexcompute/compute#4537 (squeeze frequency dimension + reshape guard in `make_eps_data`)

## Test plan
- [x] All 39 code cells execute without error
- [ ] Verify output images/plots match expected results

🤖 Generated with [Claude Code](https://claude.com/claude-code)


marc-flex (approved):
Thanks @daquinteroflex for taking care of this. I do see a few of these:
`/tmp/ipykernel_314722/3189584147.py:6: UserWarning: FigureCanvasAgg is non-interactive, and thus cannot be shown
  plt.show()`

Does that mean the image isn't displayed?

PR #469, opened 2026-04-08
Mode Simulation tutorial

FilipeFcp:
Hi guys.
I updated the Mode solver tutorial for using Mode simulation. 
Once that is good to go, I think we can use this as a reference to make AI update the other notebooks.


alec-flexcompute (approved):
Looks good, I actually didn't know about these updates. I'm glad the API has changed to this

tomflexcompute (approved):
Thanks @FilipeFcp looks great.

PR #470, opened 2026-04-10
Filipe/web run parallel

FilipeFcp:
Hi guys.
I updated another notebook. Now the parameter scan to use the new `web.run`.

I think we should also update the FAQ, but I am wondering if we will keep the `web.batch` approach, or not.
image

Should I just replace all of this, or add a new section?

### DIFF
--- a/ParameterScanWebRun.ipynb
+++ b/ParameterScanWebRun.ipynb
@@ -29,27 +29,7 @@
    "metadata": {
     "tags": []
    },
-   "outputs": [
-    {
-     "data": {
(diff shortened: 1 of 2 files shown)


alec-flexcompute (approved):
Thanks for updating this, and it looks great! (Much cleaner than our previous api)

Personally, I think we should completely replace the FAQ to adjust for this change, so count me as a vote for that.

tomflexcompute (approved):
Thanks @FilipeFcp !

PR #471, opened 2026-04-13
updated image of adjoint 27

alec-flexcompute:
added image, reference to image in preamble for the example, and updated the metadata


tomflexcompute (approved):
Thanks @alec-flexcompute . The image looks good. Just to neatpick a bit, the blue color seems to be a little lighter than our other schematics. Also can you get rid of the shadow and make the background transparent? Transparent background looks better if the user is using dark mode in the browser. 

If it's too much effort we can leave it for now.

alec-flexcompute:
should be changed now

tomflexcompute (commented):
Nice @alec-flexcompute but I don't think the image has a transparent background actually. When you ask AI to generate a transparent background, it generates a pixellated background but that's fake transparency.  The best way I found is to ask to generate a white background and then run a background remover such as this[https://www.remove.bg/] to make the background transparent. See the attached image that you can use.

Image

alec-flexcompute:
ah, nano banana 2 was being sneaky! The image should be updated now

tomflexcompute (approved):
Looks good to go. Thanks @alec-flexcompute

alec-flexcompute:
Merged! Thanks again for the background removal link - I now have it bookmarked!

PR #473, opened 2026-04-15
Update CPWRFPhotonics1 notebook after WavePort improvements

yuanshen-flexcompute:
Made changes to the wave port definition and model setup, in view of improvements to the wave port and simplification to the user experience.


mingzhou-flex (commented):
**Cell 44**: 
1. lists "n_eff, α, and α" — the third symbol should be β if I understand it correctly. 
2. Also, the formula "γ = β + jα" is not consistent with the code; code uses γ = α+ jβ

tomflexcompute (approved):

PR #477, opened 2026-04-20
updated function for fitting kwargs

alec-flexcompute:
here's the fix I made (only in cell 3), should be a quick review.


caseyflex (changes requested):
I think the peak finding is not working robustly, it seems not to correspond to the actual peaks. Gemini suggests scipy.signal.find_peaks with prominence=0.1

Based on the visual data in your plot and filtering out the numerical noise from your original arrays, both curves have exactly 5 true resonance dips within this frequency window.

Because FSR naturally changes slightly across the spectrum due to optical dispersion, it is standard practice to calculate the spacing between each adjacent peak and then take the average. Here is the corrected FSR for each simulation:

2D Simulation (Green Curve)

Filtering out the two high-frequency micro-ripples leaves the 5 actual physical resonances.

True Dips (µm): 1.59867, 1.57633, 1.55467, 1.53400, 1.51433

Individual FSR Spacings (nm): 22.33, 21.67, 20.67, 19.67

Average FSR: 21.09 nm

3D Simulation (Orange Curve)

Removing the microscopic fake dip at 1.55 µm leaves the 5 true physical resonances.

True Dips (µm): 1.58100, 1.56200, 1.54300, 1.52467, 1.50667

Individual FSR Spacings (nm): 19.00, 19.00, 18.33, 18.00

Average FSR: 18.58 nm

Note on Dispersion: As expected in real optical resonators, you can see that the individual FSR values are not perfectly constant. The spacing gradually shrinks at shorter wavelengths (higher frequencies) due to the waveguide's group velocity dispersion (GVD).

caseyflex:
@momchil-flex fyi, the fit is not great here, it should be improved with this PR

https://github.com/flexcompute/compute/pull/4795

a specialized vector fitting pole relocation for sellmeier lossless models

alec-flexcompute:
Thanks @caseyflex ! I implemented your feedback and I didn't know your note on dispersion, so I added it as a markdown cell after the FSR comparison

caseyflex (approved):

PR #480, opened 2026-04-21
fix PEC frame/PML extrusion collision in TSV

dbochkov-flexcompute:
just increased z sim size from 400 to 450


momchil-flex (approved):
Fine to get in right now, but this might be related to the issue brought up by Yangchen on slack just now - it's hard to get "correct" results, there's a lot that the user might need to tune and it's hard to pick good starting parameters.

momchil-flex:
Actually I guess it's not exactly related as here it's an actual poor interaction between separate types of boundary handling, but maybe a validator is needed then?

PR #481, opened 2026-04-21
fix title

FilipeFcp:


tomflexcompute (approved):
Maybe we can keep `ModeSimulation` in the title to highlight that it's a new class?

PR #485, opened 2026-04-23
fix the plot using new api version

FilipeFcp:
Really minor fix, just correcting the plot at cell 19


alec-flexcompute (approved):
Looks good, not sure what it was before

PR #487, opened 2026-04-24
Simplified the mesh of the Charge solver in ChargeSolver.ipymb

marc-flex:


momchil-flex:
Pretty interesting that this could work now. So only conductor-semiconductor interfaces needed extra refinement? Actually I am confused.

```
ref_line_contact_n = td.GridRefinementLine(
    r1=(w_structure / 2, h_side, 0),
    r2=(w_structure / 2 - w_contact, h_side, 0),
    dl_near=dl_boundaries,
    distance_near=dist_boundaries,
    distance_bulk=dist_bulk,
)

# mesh
mesh = td.DistanceUnstructuredGrid(
    dl_interface=dl_boundaries,
    dl_bulk=dl_bulk / 1,
    distance_interface=dist_boundaries,
    distance_bulk=dist_bulk,
    relative_min_dl=0,
    sampling=500,
    non_refined_structures=[oxide.name],
    mesh_refinements=[
        ref_region_core,
        ref_line_contact_p,
        ref_line_contact_n,
    ],
)
```

It seems like you add a refinement line at the contact interface with `dl_near=dl_boundaries`, but the grid definition already has `dl_interface=dl_boundaries`. Like, I don't see what this refinement line would add.

Do you think `ramp_up_iters=3` instead of previously 1 was related to this working now?

In the last plots, while the order of the loss curves switches between red/black 
image

it's interesting that the match is still restored at the 1575 wavelength. Actually the final plot looks suspiciously identical to the old one. Are you sure it was rerun?

`dl_boundaries = res * 0.17 * 4` this seems so arbitrary :)

marc-flex:
> It seems like you add a refinement line at the contact interface with dl_near=dl_boundaries, but the grid definition already has dl_interface=dl_boundaries. Like, I don't see what this refinement line would add.

The problem is that, currently, metal contacts are removed so the contact stops being an interface and it isn't refined unless you use the refinement region. Once we add metals to the simulation we'll be able to further remove these refinements. 

> Do you think ramp_up_iters=3 instead of previously 1 was related to this working now?

I'm not sure. It definitely didn't work now without some ramp up iters. I think we arrived at the previous mesh for different reasons:
1. when I first started dealing with this problem we didn't have refinement options so I probably worked around this with other options which lead to poor results. 
2. Vasily implemented refinements and when you suddenly have a hammer every problem becomes nail. In his defense, we kind of lead him to assume our problem was a refinement problem, which it was, though maybe in a slightly different way, as we now know. 
3. We didn't really played much with the mesh afterwards. We became complacent and assumed that's what was needed. Which was convenient since we had so many other things to do

> In the last plots, while the order of the loss curves switches between red/black

Yes. In this notebook we had the same problem Amin had: we were using oxide permittivity at low frequencies. I updated it and changed results (for the better I'd say)

> it's interesting that the match is still restored at the 1575 wavelength. Actually the final plot looks suspiciously identical to the old one. Are you sure it was rerun?

I didn't rerun that case. And in fact, I'm tempted to remove it. What do you think? Also, do you think tweaking the mode solver would improve results?

> dl_boundaries = res * 0.17 * 4 this seems so arbitrary :)

Super arbitrary. I'll combine the 0.17*4 factor and maybe add a comment like: `we modify the above defined resolution (res) to further refine the mesh at interfaces`

momchil-flex:
> > It seems like you add a refinement line at the contact interface with dl_near=dl_boundaries, but the grid definition already has dl_interface=dl_boundaries. Like, I don't see what this refinement line would add.
> 
> The problem is that, currently, metal contacts are removed so the contact stops being an interface and it isn't refined unless you use the refinement region. Once we add metals to the simulation we'll be able to further remove these refinements.

Ah, got it.

> > Do you think ramp_up_iters=3 instead of previously 1 was related to this working now?
> 
> I'm not sure. It definitely didn't work now without some ramp up iters. I think we arrived at the previous mesh for different reasons:

I think this is important to figure out before we change the notebook - and also to drive our auto mesh improvements. I wonder if the culprit is the sharp doping jump in these regions on either side of the junction:

image

Even if not, could you please try to figure out which refinements are the culprit exactly, such that removing them requires increasing `ramp_up_iters` from the default 1. We kind of don't want the user to be hitting this almost ever.

> Yes. In this notebook we had the same problem Amin had: we were using oxide permittivity at low frequencies. I updated it and changed results (for the better I'd say)
> 
> > it's interesting that the match is still restored at the 1575 wavelength. Actually the final plot looks suspiciously identical to the old one. Are you sure it was rerun?
> 
> I didn't rerun that case. And in fact, I'm tempted to remove it. What do you think? Also, do you think tweaking the mode solver would improve results?

Ah ok! Removing the last plot makes sense. Do you have any specific idea about tweaking the mode solver? I don't think it would change these results.

> > dl_boundaries = res * 0.17 * 4 this seems so arbitrary :)
> 
> Super arbitrary. I'll combine the 0.17*4 factor and maybe add a comment like: `we modify the above defined resolution (res) to further refine the mesh at interfaces`

Sounds good.

marc-flex:
> Even if not, could you please try to figure out which refinements are the culprit exactly, such that removing them requires increasing ramp_up_iters from the default 1. We kind of don't want the user to be hitting this almost ever.

It turns out we don't need ramp-up iters. Most likely I left it there because the GPU solver needed it. 

> Ah ok! Removing the last plot makes sense. Do you have any specific idea about tweaking the mode solver? I don't think it would change these results.

I didn't have any thing in mind but I was hopeful you'd have some trick.

momchil-flex (commented):
This is great! Just two comments here:

Image

- remove the `#old - mesh` comment - I don't see the purpose of it?
- having everything expressed in terms of res is great, but now res seems a bit arbitrary. Why not define it as the interface resolution to begin with so `dl_boundaries = res`?

momchil-flex (approved):

PR #493, opened 2026-05-05
fix: varactor antenna rerun

daquinteroflex:


VaractorTunedPatchAntenna.ipynb
dmarek-flex, line 1: One of the last cells has a bunch of errors, that doesnt look good

daquinteroflex: Oh no the rerun had issues then

PR #494, opened 2026-05-05
Lower shutoff and extend run_time to capture low transmission

weiliangjin2021:
Use shutoff=2e-8 (below the default) and increase run_time to 3e-11 so the FDTD simulation resolves the deep transmission dips at the Bragg resonant frequencies.


dmarek-flex (approved):
Nice, thanks for the help finding the issue! Also nice that the final flexcredit cost went down, even though the shutoff is smaller

PR #495, opened 2026-05-06
Fix varactor-tuned patch antenna notebook for 2.11.2

dmarek-flex:
Companion to #494 — fixes the varactor-tuned patch antenna notebook for the 2.11.2 release. (Vivaldi fix moved to a separate PR.)

## VaractorTunedPatchAntenna
- Switch the batch map to string task names with a separate `v1v2_by_name` lookup, and download each result to disk so the post-processing step can load them one at a time. Keeps memory bounded across the full V1/V2 sweep.
- Raise the axial-ratio plot cap from 40 dB → 100 dB so linearly polarized configurations don't appear clipped at the cap.

## Test plan
- [x] Notebook executed end-to-end on 2.11.2
- [x] Varactor sweep completes without OOM and AR plot reads correctly


dmarek-flex:
This should fix it. The Vivaldi antenna one is very sensitive to the exact PEC boundary placement, so had to add a bit more spacing now that mode solver places BCs closer to the mode plane.

daquinteroflex (approved):

PR #496, opened 2026-05-06
Update Vivaldi antenna mode solve boundaries

momchil-flex:
## Summary

This updates `IntegratedVivaldiAntenna.ipynb` for the corrected ModeSolver hard-boundary behavior. The notebook now gives the plasmonic and hybrid mode solves transverse PML with `num_pml=(10, 10)` plus a small transverse mode-plane pad so the open modes can decay without being clipped by the requested solve plane.

The phase-matching sweep is restricted to the converged overlap of the silicon and plasmonic guide sweeps. The selected target index is changed to `2.17`: this is a rounded lower-end value inside the PML-converged interpolation range, not a fine-tuned value chosen to recover the previous geometry. The resulting geometry is `w = 373.00 nm`, `p = 254.97 nm`, and `L_c = 1653.68 nm`.

The notebook was rerun end to end with this geometry. The coupling FDTD result at `1.55 um` has coupling efficiency `-1.28 dB` and return loss `-40.7 dB`. The return-loss notch shifts noticeably compared with the earlier `2.18` output, which is expected because the changed plasmonic width and coupling length change the coupler reflection interference.

The final antenna directivity peak is `13.22` (`~11.2 dB`), so the existing `ax.set_rlim(0, 15)` remains appropriate.

## Validation

- Executed the full notebook through `jupyter nbconvert --execute`.
- Coupling FDTD task: `fdve-d695af7e-de63-4dac-a9e6-4ff24a7ab674`.
- Antenna FDTD task: `fdve-342163e1-7dbb-429f-99aa-a427de558eaa`.
- Measured coupling at `1.55 um`: `T = -1.28 dB`, `R = -40.7 dB`.
- Measured antenna directivity: `D_e_max = 13.11`, `D_h_max = 13.22`.
- Converted the executed notebook with `jupytext --to py:percent`.
- Compiled the converted script with `py_compile`.


dmarek-flex:
Looks good to me, the only surprise is the resonance in S parameters is gone (or at least shifted)

image

dmarek-flex (approved):
Just check return loss in case you missed it

momchil-flex:
Yep I noticed it too but the geometry is also different so that is not too surpising! But actually I had not properly rerun that cell with the final geometry so still good you pointed it out. The resonance is back although not exactly the same as before as the geometry is still slightly different.

PR #500, opened 2026-05-07
Update PML extrusion notebook diagnostics

momchil-flex:
## Summary
- update the divergence troubleshooting notebook for default PML structure extrusion and absorber tradeoffs
- add slanted-waveguide figures showing the extruded-PML kink and reflection spectrum
- update the waveguide-to-ring notebook to compare default PML extrusion against absorber, focusing on the counter-propagating drop-mode diagnostic
- add an exploratory script for regenerating the slanted-waveguide PML/absorber comparison figures

## Validation
- ruff check DivergedFDTDSimulation.ipynb WaveguideToRingCoupling.ipynb tests/slanted_waveguide_pml_reflection.py
- nbformat validation for DivergedFDTDSimulation.ipynb and WaveguideToRingCoupling.ipynb
- Python 3.9 syntax parse for tests/slanted_waveguide_pml_reflection.py
- git diff --check
- reran WaveguideToRingCoupling.ipynb end to end with nbconvert


tomflexcompute (approved):
Thanks @momchil-flex all looks good. I wonder if it's worth it to mention this `extrude_structures` is only available for 2.11.2 and later? Just so users using earlier versions are aware they still need to do it manually or upgrade to the latest version.

momchil-flex:
> Thanks @momchil-flex all looks good. I wonder if it's worth it to mention this `extrude_structures` is only available for 2.11.2 and later? Just so users using earlier versions are aware they still need to do it manually or upgrade to the latest version.

Good idea - actually it's starting 2.11, but we just caught this regression now. Added to both notebooks!

PR #510, opened 2026-05-28
Filipe/tfln vpi fix

FilipeFcp:
Small fix that a user pointed out.

---

> [!NOTE]
> **Low Risk**
> Documentation/example-only correction to a printed figure of merit; no runtime product or auth paths affected.
> 
> **Overview**
> **Note:** The PR diff was not available in this review; this summary is inferred from the branch name (`tfln vpi fix`), description, and the **TFLN electro-optic modulator** example notebook (`TFLNTidy3d.ipynb`).
> 
> The change addresses a **user-reported error** in how **Vπ·L (push-pull)** is reported at the end of the thin-film lithium niobate modulator workflow, after voltage sweeps and effective-index extraction.
> 
> That section computes `dneff_dV` from the mode batch, defines `vpil` in **V·μm**, and prints **V·cm**. A typical fix in this pattern is correcting **double application of the push-pull half-wave factor** or the **μm → cm** scaling in the final `print` (the notebook currently uses `vpil * 1e-4 / 2` alongside a `vpil` formula that already includes a factor of ½).
> 
> Expect a **small, localized edit** to the last Vπ·L cell (and possibly aligned wording in the hybrid SOH example if the same bug was duplicated)—not simulation geometry or API changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a995aeeaccaf70c39be0d903d25d7a633c09949b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tomflexcompute (approved):
Thanks. @FilipeFcp VpiL seems quite high? Is it matching the paper better?

FilipeFcp:
@tomflexcompute This one's frustrating. The paper reports a Vπ·L of around 1.5 V·cm. I've double-checked our calculation, and had both Claude and GPT review it independently, and everything seems internally consistent. 

There is one ambiguous geometric definition, but it would not account for a gap this large.

The paper derives Vπ from an analytical formula based on the electro-optic mode overlap, whereas we compute it directly by perturbing the refractive index through the Pockels effect and re-solving for the optical mode. This is also how Photonforge does, and the results are also around 3 V.cm

tomflexcompute:
I see. Interestingly very rarely the measured result is much better than the theoretical result. 

Any way if you are confident about the result we can publish it as it is.

FilipeFcp:
It is not exactly experimental. They simulate the overlap factor and plug it into the V.pi calculation.
I will try to do the same and see if there is a better match

FilipeFcp:
I tried calculating the same way the paper does, and sweeping some geometric parameters. Still cant get closer to paper value. Maybe there is some that is not reported.

PR #530, opened 2026-06-26
feat: add autograd clip notebook example

groberts-flex:
This example notebook demonstrates autograd through clipped geometries to make a light extraction structure! It uses clip to model a fabrication process and take gradients through it directly.

---

> [!NOTE]
> **Low Risk**
> Documentation-only addition (example notebook and images); no library or production code paths change.
> 
> **Overview**
> Adds **`Autograd32ClipOperationLightExtractor.ipynb`**, a new inverse-design example that walks through remote adjoint optimization when the etched surface is built from **boolean `ClipOperation` geometries** (two hexagonal hole lattices combined, default **intersection**, clipped to a finite pixel).
> 
> The notebook optimizes pitches, radii, and relative rotation for upward flux via **`autograd.value_and_grad`** and Tidy3D’s Adam helper, then extends the story with a **fabrication cleanup** step on sub-threshold features and **top LEE** comparisons (initial vs optimized, raw vs processed, vs a flat reference), including polarization and dipole-position averaging via **`web.Batch`**. Supporting assets are referenced under `./img/` (thumbnail and lithography-process schematic), and notebook metadata points at **`Autograd32ClipOperationLightExtractor.qmd`** for Quarto publishing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c8c7696fe179729b83daf83781bc4d6c0035858. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


marcorudolphflex:
Cool work, great feature!
All in all, I would just make it much shorter (mostly by cleaning code and docs).

Also some formalities:
- point this branch to pre/v2.12
- ensure that the metadata are complete (feature image, title, description, keyowrds)
- fix ruff linting
- ensure dictionary is complete (see spell checks)
- ensure entry in `docs/features/autograd.rst` 

Regarding the notebook itself:
- avoid autograd32 naming of tasks to be robust against future migrations and not confuse non-repo users
- would remove the output cell containing `WARNING: Using canonical configuration directory `
- The notebook's first sentence comes with "autograd support with boolean ClipOperation geometries". I am not sure if any reader understand the term `ClipOperation` here? Maybe better explain that more understandable?
- not sure if this is that important to mention in the setup? "The optimizer is Tidy3D's built-in Adam helper from tidy3d.plugins.autograd." - I think this is later repeated anyways
- consider some explanatory comments or named subsections in the cell which defines constants (starting with `air_index = 1.0`)
- consider removal or shortage of some docstrings. For example, I wouldn't have more than 1 line docstring for `format_params` or `project_params`
- is there a reason for a separate cell containing `operation = "intersection"`?
- The notebook could be more compact. Do we need an own function for `build_interface_reference_structure`, `build_static_gan_structures`, `build_lateral_source_positions` or `build_traced_etch_structure`, they are called once in the entire notebook
- "Adam minimizes by default, so we pass the negative gradient to maximize the upward extracted flux." Note that our Adam has a direction arg which can be set to `max` here
- let's make use of the adjoint-capable `FluxMonitor`
- make introductory image bigger
- do we want to print the first `value_and_grad = ag.value_and_grad(objective)` - or should it be more of a smoke check? 
- `build_optimization_simulation` exposes `operation`, `source_center`, `source_polarization`, and `grid_spec`, but the optimization path always uses the same values. We could just simplify to `params`. Also check other arguments which are either constants or use defaults which are never excited like `monitors=None` in `build_simulation`
- `top_spectrum` is never used? `calculate_top_lee` may be expanded inside `summarize_lee_result`
- not sure if this is confusing for users: "No Flexcredit cost for the whole batch as all simulations were restored from local cache." Could someone misunderstand this and assume a free simulation?
- "Our figure of merit is the frequency-averaged upward flux"  - but `monitor_freqs` is just a single freq0?
- `cleaned_etch_multipolygon` and `cleaned_etch_geometry` are not used?
- I think the md text in "Fabrication Cleanup Prototype" should mention that we went for `fabrication_cleanup_mode = "open_close"`. Also I wonder if we need to show the other variants like open/close or area only when we go for one in the end and do not really evaluate others? Would make things shorter and put more focus on the rest.
- "we sample a small lateral grid spanning one wavelength in the x and y directions", isn't it 2 wavelengths having [-0.5, 0, 0.5] -> total span of 1 um = 2 * 0.5 um?
- can't we just merge the polarization-resolved center dipole extraction and center dipole extraction plot? (cell 25/27). Would make things shorter and less redundant.
- I had some visual analytics course a couple of years ago and they told us to never connect points if there is not really an intuition for an interpolation which is the case for the "position-dependent top extraction" - can't we just make 3x3 bar plots? This would also make the introduction of source position indices less important - We could think about dropping the table.

groberts-flex:
thank you for this great feedback! I addressed these points and here a few responses to some of them as well with some more detail:

"The notebook could be more compact. Do we need an own function for build_interface_reference_structure, build_static_gan_structures, build_lateral_source_positions or build_traced_etch_structure, they are called once in the entire notebook"

Inlined build_interface_reference_structure() inside build_interface_reference_simulation()
Inlined build_lateral_source_positions() in the source-position section.

Kept build_static_gan_structures() and build_traced_etch_structure() as top-level helpers but moved to live immediately above build_simulation() where they are called


"do we want to print the first value_and_grad = ag.value_and_grad(objective) - or should it be more of a smoke check?"

we do this smoke check now

sim0 = build_optimization_simulation(params0)
sim0.validate_pre_upload()


I moved the PR over to compute and updated the name of the notebook to match the current numbering

